PR #40, opened 2014-11-26
fix multi-line DEBUG_TRACE macro calls

erantapaa:
Using GHC 7.8.3 (e.g. under the Haskell Platform 2014.2.0.0) the template file `GLR_Lib` is created with two indentation errors:
- line 161, the line `case new_stks of` begins with a space and not a tab
- line 190, the line `stks' <- foldM (pack i) stks reds` begins with a space and not a tab

(These line numbers reference the files generated by `happy-1.19.4`)

These indentation errors will cause compile errors when happy is used with the `--glr` flag. For instance, see this StackOverflow question: http://stackoverflow.com/q/27150105/866915

The problem seems to be due to the multi-line macro call just before these lines. Changing them to a single line fixes the problem. For some reason `ghc -E -cpp` is replacing the leading tab on these lines with a space.

### DIFF
--- a/templates/GLR_Lib.hs
+++ b/templates/GLR_Lib.hs
@@ -178,8 +178,7 @@ doActions stks (tok:toks)
                              shiftAll tok_form stks'
                          | tok_form <- tok ]
 	let new_stks = merge $ concat stkss
-	DEBUG_TRACE(unlines $ ("Stacks after R*/S pass" ++ show tok)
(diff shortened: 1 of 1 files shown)


erikd:
Are you by any chance using Clang as the C compiler?

simonmar:
Old pull request, please rebase and resubmit if this is still a problem.

PR #49, opened 2015-07-16
Add signature generation for using typeclasses in parsers

emc2:
This pull requests causes happy to generate some new signatures that allow typeclasses to be used in monadic parsers.  This makes no changes to actual code; it only generates some signatures.

This patch only generates signature for parsers that combine `%monad` and `%lexer` directives.  There appear to be some non-trivial issues that arise when `%monad` is used without `%lexer`.

This has been verified to work with a complex parser (and in conjunction with the corresponding alex pull request) here:

https://github.com/saltlang/saltlang/blob/typeclass_parser/src/salt/Language/Salt/Surface/Parser.y

The resulting lexer/parser combination operates inside a monad completely described by typeclasses.

### DIFF
--- a/src/ProduceCode.lhs
+++ b/src/ProduceCode.lhs
@@ -912,7 +912,8 @@ directive determins the API of the provided function.
 >       . maybe_tks
 >       . str " = "
 >       . str unmonad
->       . str "happyThen (happyParse "
+>       . str "happySomeParser where\n"
+>       . str " happySomeParser = happyThen (happyParse "
 >       . case target of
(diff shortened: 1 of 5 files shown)


erikd:

emc2:
Rebased to head

emc2:
I've added tests, including an example of the use case for monadic-typeclass parsers.  Additionally, the patch works in conjunction with the recent addition of typeclass-based lexers to alex (see the link above).

Given that alex now has typeclass support in the main release, there is a case to integrate this PR.

simonmar:
Seems reasonable...

PR #51, opened 2015-08-01
Type signature on parameterized rules

PhilThomas:
This pull request adds the ability to create type signatures for parameterized rules.  This, along with a slight change to the way that parameterized rule instance names are generated allows parameterized rules to be used with GLR parsers.

It works by substituting the formal parameter names in parameterized rule type signatures with the types associated with the actual parameters.  Only parameter names that start with a lowercase letter are substituted.  Actual parameters that correspond with non-terminals are substituted with the non-terminal's type (which must be defined).  Actual parameters that correspond with terminals are substituted with the token type, which will not work correctly when the terminal uses a `$$` in its token pattern.  For this to work, there would need to be a way to define the type of the `$$`.  A work around would be to define a non-terminal with a single production corresponding to the `$$` terminal, and to use it (and its associated type signature) instead.

In addition, parameterized rule instance names are generated by interleaving (and terminating) the actual parameters with double underscores instead of commas and parentheses so that they form valid data constructor names in the `GSymbol` data type.  Note that string and character actuals are not currently handled, and will generate illegal names if used.  There is a possibility that these auto-generated names could alias, since a non-parameterized non-terminal could be given such a name; however, the fact that the names end with a double underscore makes this unlikely.

For instance, the following grammar:

```happy
{
{-# LANGUAGE FlexibleInstances #-}
}

%tokentype { Token }
%token
  Int   { TokInt }
  Ident { TokIdent }

%%

top :: { ([Token], [Token], Maybe Token) }
    : many(Int) many1(Ident) opt(Int) { ($1, $2, $3) }

opt(p) :: { Maybe p }
       : p { Just $1 }
       | {- empty -} { Nothing }

many(p) :: { [p] }
	: many1rev(p) { reverse $1 }
	| {- empty -} { [] }

many1(p) :: { [p] }
	 : many1rev(p) { reverse $1 }

many1rev(p) :: { [p] }
	    : p { [$1] }
	    | many1rev(p) p { ($2:$1) }

{
data Token = TokInt | TokIdent
  deriving ( Eq, Ord, Show )
}
```
 Generates the following definitions for `GSymbol` and `GSem` when run with --glr --decode:

```haskell
data GSymbol = HappyEOF | HappyTok {-!Int-} (Token) | G_top 
 | G_many__Int__ 
 | G_many1__Ident__ 
 | G_opt__Int__ 
 | G_many1rev__Ident__ 
 | G_many1rev__Int__ 
   deriving (Show,Eq,Ord)

data GSem
 = NoSem
 | SemTok (Token)
 | Sem_0 ([(Token)] -> [(Token)] -> Maybe (Token) -> ([Token], [Token], Maybe Token)) 
 | Sem_1 ([(Token)] -> [(Token)]) 
 | Sem_2 ([(Token)]) 
 | Sem_3 (Token -> Maybe (Token)) 
 | Sem_4 (Maybe (Token)) 
 | Sem_5 (Token -> [(Token)]) 
 | Sem_6 ([(Token)] -> Token -> [(Token)]) 
```

### DIFF
--- a/src/Grammar.lhs
+++ b/src/Grammar.lhs
@@ -297,8 +297,35 @@ Translate the rules from string to name-based.
 >   rules2 <- mapM transRule rules1
 
 >   let
->       tys = accumArray (\_ x -> x) Nothing (first_nt, last_nt)
->                       [ (nm, Just ty) | (nm, _, Just ty) <- rules1 ]
+>       type_env = [(nt, t) | (nt, _, Just (t,[])) <- rules] ++
+>                  [(nt, getTokenType dirs) | nt <- terminal_strs] -- XXX: Doesn't handle $$ type!
+>
+>       fixType (ty,s) = go "" ty
+>         where go acc [] = return (reverse acc)
+>               go acc (c:r) | isLower c = -- look for a run of alphanumerics starting with a lower case letter
+>                                let (cs,r1) = span isAlphaNum r
+>                                    go1 x = go (reverse x ++ acc) r1
+>                                in case lookup (c:cs) s of
+>                                        Nothing -> go1 (c:cs) -- no binding found
+>                                        Just a -> case lookup a type_env of
+>                                          Nothing -> do
+>                                            addErr ("Parameterized rule argument '" ++ a ++ "' does not have type")
+>                                            go1 (c:cs)
+>                                          Just t -> go1 $ "(" ++ t ++ ")"
+>                            | otherwise = go (c:acc) r
+>
+>       convType (nm, t)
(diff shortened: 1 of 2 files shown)


PhilThomas:
Note that this does not (yet) support nested applications of parameterized rules, such as `opt(many1(Identity))`.  To handle this, the `type_env` would need to be extended with the substituted types, and `fixType` would need to recurse to perform substitutions in arguments themselves (and be able to detect cycles in the dependency graph).

simonmar:
Thanks!

PR #52, opened 2015-08-11
Type signatures for nested parameterized rules

PhilThomas:
This pull request adds support for type signatures on *nested* parameterized rules.

Since type signatures can now depend on the types of other rules, it first builds an array of `ShowS` functions that generate typenames.  The array is built in the M monad, so that errors can be reported.  Then, a new array of actual type names is built by applying the `ShowS` functions to the array we are constructing (tying a knot)!  The type dependencies should form a DAG, and not a general graph, so no cycles can occur, and the algorithm will terminate.  The new array maps both terminals and non-terminals to their respective types; however, terminals are always mapped to the token type, so any token that uses a `$$` will yield the wrong type.


erikd:

PhilThomas:
Very cool!  I took the opportunity to clean up a little too.  Should be good to go now.

PhilThomas:
I found an issue with `ProduceGLRCode.mkDecodeUtils`.  The types of two elements in `seminfo` may appear to be different when compared as `String`s, even though they may be equivalent to Haskell.  For instance `[Maybe (Token)]` and `[Maybe Token]` are not `String` equal, even though they represent equivalent types.  This causes separate instances of `TreeDecode` for equivalent types to be created, making ghc very unhappy.  I suppose that this was always a potential problem, but previously, the user was in complete control of the `String` form of all types, and could be counted upon to make sure that equivalent types had the same representation.  However, when types are generated for parameterized rules, additional parentheses are added in case they are needed, causing the `String` representations to mismatch.  Some kind of normalization is called for.  Maybe it would be sufficient to only add parens when absolutely necessary.

PR #57, opened 2016-02-10
Fix Generalized LR Parsing with newer GHC

thomie:
Tested with GHC 7.8.4, 7.10.3 and 8.0.1.

AMP + other fixes. I also did some cleanups in the examples/ and tests/ directory.


simonmar:
Thanks!

PR #58, opened 2016-03-11
Fix spello in Calc.ly

GetContented:
Slight spelling mistake fix (declair / declare)

### DIFF
--- a/examples/Calc.ly
+++ b/examples/Calc.ly
@@ -5,7 +5,7 @@
 
 First thing to declare is the name of your parser,
 and the type of the tokens the parser reads.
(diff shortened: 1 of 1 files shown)


examples/Calc.ly
```diff
@@ -5,7 +5,7 @@
 
 First thing to declare is the name of your parser,
 and the type of the tokens the parser reads.
-
+d
```
simonmar: spurious change?

GetContented: Accidental, actually. I think I deleted this PR didn't I?

PR #65, opened 2016-05-16
Add a missing import following #64

ndmitchell:
Sorry Simon, I missed an import from https://github.com/simonmar/happy/pull/64 :frowning:


ndmitchell:
I double tested this one in a clean repo, so fairly certain I didn't mess this one up...

PR #66, opened 2016-05-16
Rewrite fold_lookahead to avoid using the stack and improve time complexity

ndmitchell:
As part of my efforts to investigate space leaks, I ended up at the `fold_lookahead` function, which I rewrote as below. I offer this as an entirely separate patch because there are good reasons you might want to not merge it. Advantages:

* Reduces stack usage from > 1Mb to < 1Kb.
* Algorithm more directly states what it does.
* Better time complexity - previously it was a combination of insertion sort on the Int and a weird order-changing nub construction on the Lr0Item.
* Passes all the tests.

Disadvantages:

* Has no meaningful time difference - my tests are with 0.03s.
* Produces different output since the list is in a different order. I believe the results are equivalent, but you will need to say for sure.

### DIFF
--- a/src/LALR.lhs
+++ b/src/LALR.lhs
@@ -22,7 +22,8 @@ Generation of LALR parsing tables.
 > import Control.Monad.ST
 > import Data.Array.ST
 > import Data.Array as Array
-> import Data.List (nub,foldl')
+> import Data.List (nub,foldl',groupBy,sortBy)
+> import Data.Function (on)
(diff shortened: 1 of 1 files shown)


simonmar:
The patch looks sensible to me.  This codebase is very very very old you know :)

ndmitchell:
I was impressed just quite how many `foldr`s I found in the code base :smile:.

simonmar:
I guess I was going through a foldr-happy phase.

PR #68, opened 2016-05-26
GenericTemplate.hs: workaround GHC 7.6 type checking failure

da-x:
Even with an exact type signature for `unbox_int `and `indexShortOffAddr` GHC 7.6 fails to realize the type for the second parameter of indexShortOffAddr which needs to be of unboxed kind '#' instead
of boxed '*'.

```
    Couldn't match kind `#' against `*'
    In the second argument of `indexShortOffAddr', namely
      `(unbox_int $ bit `div` 16)'
```

It's probably due to `div`. Figured out we can use unboxed right shift instead, to get it straight.

Tested with GHC 7.6.3, GHC 7.8.4, and GHC 7.10.3.

Signed-off-by: Dan Aloni <dan@kernelim.com>

### DIFF
--- a/templates/GenericTemplate.hs
+++ b/templates/GenericTemplate.hs
@@ -151,12 +151,13 @@ indexShortOffAddr arr off = arr Happy_Data_Array.! off
 #endif
 
 
-readArrayBit arr bit =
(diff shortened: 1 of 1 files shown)


simonmar:
Thanks for jumping on this so quickly!

PR #69, opened 2016-06-15
GLR changes

mrd:
Gathering several changes that have been sitting on my computer for too long. These fix some outstanding annoyances in the --glr mode, including: tabs, deprecated command line options, a broken debug mode, and the inability to select a specific module name for a grammar.


erikd:

PR #74, opened 2016-09-01
Bunch of minor fixes

erikd:
Biggest change is the .travis.yml file which tests with the last 4 major releases of GHC.


simonmar:
Thanks!

src/ProduceCode.lhs
```diff
@@ -1319,7 +1320,7 @@ slot is free or not.
 > hexChars acts = concat (map hexChar acts)
 
 > hexChar :: Int -> String
-> hexChar i | i < 0 = hexChar (i + 2^16)
+> hexChar i | i < 0 = hexChar (i + 65536)
```
int-index: How's this an improvement?

erikd: It prevents a compiler warning (type defaulting due to `^`). Most software engineers should recognise `65536` as `1 << 17` or `2^16`.

PhilThomas: Yes, one or the other, but not both ;-)

PR #77, opened 2016-10-25
Add an option to pretty print just the grammar rules to a file.

yav:
This pretty prints the original input file in a form where all the semantic
actions and type signatures are omitted.  It is useful when one is trying
to understand just the structure of a complicated grammar.


happy.cabal
```diff
@@ -1,5 +1,5 @@
 name: happy
-version: 1.19.6
+version: 1.19.7
```
simonmar: Let's put this in a separate commit, we might include several changes in a release.

yav:
OK, I updated the cabal file back to the old version.

PR #79, opened 2016-12-05
Fix broken command line options.

yav:
The commit adding the pretty printing functionality accidentally
broke the `-i` flag.   `-p` was also broken, in that it would print
to `stdout` all the time.

Should be fixed now, sorry for the breakage.

### DIFF
--- a/src/Main.lhs
+++ b/src/Main.lhs
@@ -164,14 +164,32 @@ Report any conflicts in the grammar.
 
 Print out the info file.
 
+>       getInfoFileName name cli                >>= \info_filename ->
+>       let info = genInfoFile
+>                       (map fst sets)
+>                       g
+>                       action
+>                       goto
+>                       (token_specs g)
+>                       conflictArray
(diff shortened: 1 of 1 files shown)


erikd:
Ok, that fails to build in travis because the cabal install location changed. I'll whip up a fix for that.

erikd:
PR #80 should fix the travis build. Once the fix is in master you can rebase and then this should build.

erikd:
@yav Is it possible to write a simple test for this?

yav:
I am not sure, happy's tests appear to just `diff` against `stdout`, while to test `-i` and `-p` with have to compare the contents of generated files, which---I think---requires changes to the test frame-work.  Let me know if I am wrong.

The thing where I accidentally printed a whole bunch of cruft to `stdout` should be detected by one of the tests that compares against an expected output, so we should be covered there, no need for new tests I think.

erikd:
Fair enough!

PR #92, opened 2017-08-02
'happyNewToken' now uses the right token type

### LINKED ISSUE #91: Generated signature for `happyNewToken` ignores `%tokentype`
When generating code for `happyNewToken`, Happy does not use the token type given by `%tokentype`, hardcoding instead to `Token`. [Here is the relevant code](https://github.com/simonmar/happy/blob/41fbb827d6fb1761233b015382fd62808abe45cc/src/ProduceCode.lhs#L433-L447):

               str "happyNewToken :: " . pcont . str " => "
             . str "(Happy_GHC_Exts.Int#\n"
             . str "                   -> Happy_GHC_Exts.Int#\n"
             . str "                   -> Token\n"
             . str "                   -> HappyState Token (t -> "
             . pty . str " a)\n"
             . str "                   -> [HappyState Token (t -> "
             . pty . str " a)]\n"
             . str "                   -> t\n"
             . str "                   -> " . pty . str " a)\n"
             . str "                 -> [HappyState Token (t -> "
             . pty . str " a)]\n"
             . str "                 -> t\n"
             . str "                 -> " . pty . str " a\n"

I've already fixed this issue on a local branch of mine, so I'll open a PR soon.

harpocrates:
This should fix #91.

This bug can be triggered by compiling the generated code on any grammar that

  * uses `%tokentype` that is _not_ just `Token` (try something like `Token1`)
  * makes use of `%lexer`
  * is compiled using `-g`

### DIFF
--- a/src/ProduceCode.lhs
+++ b/src/ProduceCode.lhs
@@ -429,18 +429,18 @@ The token conversion function.
 >       case (target, ghc) of
 >          (TargetHaskell, True) ->
 >             let pcont = str monad_context
->                 pty = str monad_tycon  in
+>                 pty = str monad_tycon in
 >                 str "happyNewToken :: " . pcont . str " => "
 >               . str "(Happy_GHC_Exts.Int#\n"
 >               . str "                   -> Happy_GHC_Exts.Int#\n"
->               . str "                   -> Token\n"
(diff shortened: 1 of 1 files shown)


simonmar:
Thanks, any chance you could make a regression test for this?

PR #96, opened 2017-08-16
Added a regression test for #91

harpocrates:
I don't have the right soup to test this (I don't have any global version of `ghc` or `cabal` installed, I get by with `stack`) but the CI should tell us whether I correctly added a test case.


simonmar:
Thanks!

PR #98, opened 2017-08-21
Happy overflow bug

### LINKED ISSUE #93: Code generated is unpredictably incorrect
I've been running across what I think is the same bug several times in the last couple of months while working on a pretty big grammar. Here are some more extreme things I've witnessed that seem to indicate the code generated is incorrect:

  1. moving a production (for example to the top of the file) changes the behaviour of the generated code
  2. the presence or absence of a production (that is used nowhere!) changes the behaviour of the generated code

I've made a branch [here][0] which is as reduced an example as I've managed to get. There is one test suite included which has exactly one test case. Right now, I think the code generated by Happy (1.19.5 and HEAD) is incorrect (the test should pass). However, commenting out the `pat_slice` production (which is used nowhere) makes the test pass again. In fact, as I've tried to reduce this grammar over the past weeks, I'm now finding it difficult to comment out _anything_ without the test passing again.

I realize this is a huge test case, but I am at a complete loss at what to do to reduce the grammar since doing so seems to remove the bug.

[0]: https://github.com/harpocrates/language-rust/blob/happy-bug-incorrect-code/src/Language/Rust/Parser/Internal.y

harpocrates:
This fixes #93. In a nutshell:

My `happyTable`/`happyCheck` have over `2^15` entries in them. That isn't a problem by itself, if it weren't for the fact that `happyActOffsets`/`happyGotoOffsets` store indices pointing into `happyTable` in signed 16 bit numbers. Then, the issue is that the indices that are being written into `happyActOffsets` sometimes get too big. When that happens, they wrap around to the vicinity of -32000.

The fix is simple: keep track of the smallest value in `happyActOffsets`/`happyGotoOffsets`. Then, when reading values back out of these arrays, we check if the value read is smaller than what is supposed to be the smallest value. If it is, we know that there was an accidental overflow (and we add `2^16` to the value we got out).

This lets us have `happyTable`/`happyCheck` tables with a little under 2^16 entries - double what we currently have and with no extra space cost. (And enough to fix my [project](https://github.com/harpocrates/language-rust).)

### DIFF
--- a/templates/GenericTemplate.hs
+++ b/templates/GenericTemplate.hs
@@ -128,7 +128,8 @@ happyDoAction i tk st
                                      happyShift new_state i tk st
                                      where new_state = MINUS(n,(ILIT(1) :: FAST_INT))
    where off    = indexShortOffAddr happyActOffsets st
-         off_i  = PLUS(off,i)
+         !IBOX(min_off) = happyMinOffset
+         off_i = if LT(off, min_off) then PLUS(off, PLUS(i, ILIT(65536))) else PLUS(off, i)
          check  = if GTE(off_i,(ILIT(0) :: FAST_INT))
                   then EQ(indexShortOffAddr happyCheck off_i, i)
                   else False
@@ -139,13 +140,7 @@ happyDoAction i tk st
 #endif /* HAPPY_ARRAY */
 
 #ifdef HAPPY_GHC
-indexShortOffAddr (HappyA# arr) off = i_adjusted
-  where
-        !(_, Happy_GHC_Exts.I# n_rules) = Happy_Data_Array.bounds happyReduceArr
-        i = Happy_GHC_Exts.indexInt16OffAddr# arr off
-        i_adjusted = if Happy_GHC_Exts.isTrue# (Happy_GHC_Exts.negateInt# n_rules Happy_GHC_Exts.>=# i)
-                       then i Happy_GHC_Exts.+# 65536#
-                       else i
+indexShortOffAddr (HappyA# arr) off = Happy_GHC_Exts.indexInt16OffAddr# arr off
 #else
 indexShortOffAddr arr off = arr Happy_Data_Array.! off
 #endif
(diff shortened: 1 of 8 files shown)


harpocrates:
I think this is now ready to merge. When run with `--ghc`, Happy now
  
  * supports larger grammars than before (`happyTable`/`happyCheck` can be almost twice as big) without changing the underlying 16-bit byte array representation
  * fails with an error message if `happyTable`/`happyCheck` are _too_ big (the current behavior is to silently generate bogus code)

simonmar:
Good diagnosis!

Perhaps I'm missing something here, but wouldn't an easier fix be to use unsigned 16-bit offsets rather than signed offsets?

harpocrates:
I'm not sure that would work so easily. The problem is that entries in `happyActOffsets`/`happyGotoOffsets` can also be slightly negative and still meaningful. After retrieving an offset from one of these tables, the offset is added to the token number and the _sum_ of those (if positive) is used to look into `happyTable`/`happyCheck`.

All that to say that small negative offsets _are_ still used, so we can't just switch to unsigned offsets.

Does this make sense?

templates/GenericTemplate.hs
```diff
@@ -139,22 +139,16 @@ happyDoAction i tk st
...
 
 #ifdef HAPPY_GHC
-indexShortOffAddr (HappyA# arr) off =
-        Happy_GHC_Exts.narrow16Int# i
-  where
-        i = Happy_GHC_Exts.word2Int# (Happy_GHC_Exts.or# (Happy_GHC_Exts.uncheckedShiftL# high 8#) low)
-        high = Happy_GHC_Exts.int2Word# (Happy_GHC_Exts.ord# (Happy_GHC_Exts.indexCharOffAddr# arr (off' Happy_GHC_Exts.+# 1#)))
-        low  = Happy_GHC_Exts.int2Word# (Happy_GHC_Exts.ord# (Happy_GHC_Exts.indexCharOffAddr# arr off'))
-        off' = off Happy_GHC_Exts.*# 2#
+indexShortOffAddr (HappyA# arr) off = Happy_GHC_Exts.indexInt16OffAddr# arr off
```
simonmar: Why is this changed? I believe the code here is to make the index endian-independent.

harpocrates: I've changed this back - your comment is obvious in hindsight.

harpocrates:
Would you prefer if I opened a new PR on a new branch with a cleaner history? I think I mistakenly branched off of https://github.com/simonmar/happy/pull/97 instead.

simonmar:
Up to you whether you force-push to this branch or open a new PR.

harpocrates:
I've opened a new PR, so I'm closing this one.

PR #99, opened 2017-08-28
Add docbook xml path as currently found in Arch linux.

nochiel:
Allow Arch's makepkg to build and install documentation as expected.


simonmar:
Thanks!

PR #102, opened 2017-09-05
Support larger offset tables

harpocrates:
### This is an updated version of https://github.com/simonmar/happy/pull/98.

This fixes https://github.com/simonmar/happy/issues/93.

* Added new variable `happyMinOffset` that is present only when arrays are enabled.
This variable keeps track of the lowest offset value we should ever find in the
values of `happyActOffsets` and `happyGotoOffsets`. That way, if we ever find a
value that is smaller, we know it accidentally wrapped around (and we can correct
for the wraparound).

* Added a check so that if Happy is run with `--ghc` and the offset tables are
actually too large, Happy will fail with an error message instead of generating
bogus code (as it currently does).

Since the negative numbers in offset tables are generally quite small, this
effectively doubles the size of offsets supported.

### DIFF
--- a/templates/GenericTemplate.hs
+++ b/templates/GenericTemplate.hs
@@ -128,7 +128,7 @@ happyDoAction i tk st
                                      happyShift new_state i tk st
                                      where new_state = MINUS(n,(ILIT(1) :: FAST_INT))
    where off    = indexShortOffAddr happyActOffsets st
-         off_i  = PLUS(off,i)
+         off_i  = if LT(off, (unbox_int happyMinOffset)) then PLUS(off, PLUS(i, ILIT(65536))) else PLUS(off, i)
          check  = if GTE(off_i,(ILIT(0) :: FAST_INT))
                   then EQ(indexShortOffAddr happyCheck off_i, i)
                   else False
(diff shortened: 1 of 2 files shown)


templates/GenericTemplate.hs
```diff
@@ -239,7 +239,7 @@ happyMonad2Reduce k nt fn j tk st sts stk =
          let drop_stk = happyDropStk k stk
 #if defined(HAPPY_ARRAY)
              off = indexShortOffAddr happyGotoOffsets st1
-             off_i = PLUS(off,nt)
+             off_i = if LT(off, (unbox_int happyMinOffset)) then PLUS(off, PLUS(nt, ILIT(65536))) else PLUS(off, nt)
```
simonmar: It would be really nice if we could only omit this conditional if we actually need it.  Happy could figure out whether the grammar requires overflow and `#define` something that would get checked here.

Not essential, but I'm always a bit paranoid about adding some extra overhead into the central part of the algorithm like this.

simonmar:
I'll merge this as is, but it would be great if we could have a test too.

templates/GenericTemplate.hs
harpocrates: I agree. I'll work on that when I have some time (probably this coming weekend).

harpocrates:
Regarding having a test, I reckon it would have to be a pretty giant test case (at the minimum probably a thousand or so lines). Would that still be worth it? I can take a crack at this - while working on this PR I tried and failed to make a more minimal test case.

simonmar:
Even if it's a huge test case I think it's worth it.  I can imagine this functionality easily breaking in the future if we don't have a test for it.

harpocrates:
That's a very good point. I guess I'll work on extracting a standalone test case.

PR #105, opened 2017-09-13
Happy overflow followup

harpocrates:
This addresses the comments on https://github.com/simonmar/happy/pull/102. It adds

  * a (rather long, but still one file) test case which results in a `happyTable` with more than 2^15 entries. This test case fails before the PR and passes afterwards.

  * instead of using `happyMinOffset`, we generate `happyAdjustOffset`, which contains the logic for adjusting offsets only when `--ghc` is enabled _and_ the `happyTable` is more than 2^15 elements long. Otherwise, it does nothing.

Please let me know if anything is amiss. I hope this ties up loose ends from the last PR.

### DIFF
--- a/templates/GenericTemplate.hs
+++ b/templates/GenericTemplate.hs
@@ -150,15 +150,13 @@ indexShortOffAddr (HappyA# arr) off =
 indexShortOffAddr arr off = arr Happy_Data_Array.! off
 #endif
 
-{-# INLINE unbox_int #-}
-unbox_int IBOX(x) = x
-
-{-# INLINE lt #-}
-lt x y = LT(x,y)
+{-# INLINE happyLt #-}
+happyLt x y = LT(x,y)
 
 #ifdef HAPPY_GHC
 readArrayBit arr bit =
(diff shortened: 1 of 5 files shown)


templates/GenericTemplate.hs
```diff
@@ -150,8 +150,12 @@ indexShortOffAddr (HappyA# arr) off =
 indexShortOffAddr arr off = arr Happy_Data_Array.! off
 #endif
 
+{-# INLINE unbox_int #-}
 unbox_int IBOX(x) = x
```
simonmar: These functions share a namespace with the user's code, so let's use names prefixed with `happy` to avoid clashes. (I know we don't do this consistently, but we should)

templates/GenericTemplate.hs
```diff
@@ -150,8 +150,12 @@ indexShortOffAddr (HappyA# arr) off =
...
 
+{-# INLINE lt #-}
+lt x y = LT(x,y)
```
simonmar: ditto

harpocrates:
Not at all. Thanks for maintaining this even after so many years!

I moved `unbox_int` back into `readArrayBit` (no point in polluting the global namespace now that I'm not using it anywhere else) and renamed `lt` -> `happyLt`.

PR #106, opened 2017-09-15
Issue 94

### LINKED ISSUE #94: Monadic grammar, threaded lexer, `-a` enabled, but no type signature, fails to compile
I noticed that when I have a monadic grammar with a threaded lexer, generated with ~~(just)~~ `-a` enabled, and do not include a type signature on a production, I get type errors in the generated code. I don’t think this is the expected behaviour. Here is an example of such a grammar:

```
-- compileBug.y
%name parse prod

%tokentype { Token }

%monad { P } { bindP } { returnP }
%error { error "parse error" }
%lexer { lexer } { EOF }

%token
  IDENT  { Identifier $$ }

%%

-- Uncommenting the type signature below makes the generated code compile
prod -- :: { () }
  : IDENT { () }

{
data Token = EOF | Identifier String

type P a = String -> (a, String)

bindP :: P a -> (a -> P b) -> P b
bindP p f s = let (x,s') = p s in f x s'

returnP :: a -> P a
returnP = (,)

lexer :: (Token -> P a) -> P a
lexer cont s = cont (case s of { "" -> EOF; _ -> Identifier s }) ""

main = return ()
 }
```

Without the type signature:

    $ happy -a -o compileBug.hs compileBug.y
    $ ghc compileBug.hs
    [1 of 1] Compiling Main             ( compileBug.hs, compileBug.o )
    
    compileBug.hs:61:17: error:
        • Couldn't match type ‘t4’ with ‘()’
          ‘t4’ is a rigid type variable bound by
            the type signature for:
              happyReduce_1 :: forall t4.
                               Int
                               -> Token
                               -> Int
                               -> Happy_IntList
                               -> HappyStk (HappyAbsSyn t4)
                               -> P (HappyAbsSyn t4)
            at compileBug.hs:60:18
          Expected type: Int
                         -> Token
                         -> Int
                         -> Happy_IntList
                         -> HappyStk (HappyAbsSyn t4)
                         -> P (HappyAbsSyn t4)
            Actual type: Int
                         -> Token
                         -> Int
                         -> Happy_IntList
                         -> HappyStk (HappyAbsSyn ())
                         -> P (HappyAbsSyn ())
        • In the expression: happySpecReduce_1 0 happyReduction_1
          In an equation for ‘happyReduce_1’:
              happyReduce_1 = happySpecReduce_1 0 happyReduction_1
        • Relevant bindings include
            happyReduce_1 :: Int
                             -> Token
                             -> Int
                             -> Happy_IntList
                             -> HappyStk (HappyAbsSyn t4)
                             -> P (HappyAbsSyn t4)
              (bound at compileBug.hs:61:1)
    $

With the type signature uncommented:

    $ happy -a -o compileBug.hs compileBug.y
    $ ghc compileBug.hs
    [1 of 1] Compiling Main             ( compileBug.hs, compileBug.o )
    Linking compileBug ...
    $

Note that this is a regression (I have at least some older version of Happy on which the type signature is not required to generate compiling code). This bug seems present in at least ~~1.19.5 and~~ HEAD.

harpocrates:
Implements both points outlined in comment <https://github.com/simonmar/happy/issues/94#issuecomment-329585023>. Also added a new test case which didn't use to pass, but does now.

This fixes #94. Someone should probably still update the user guide to summarize what <https://github.com/simonmar/happy/pull/49> added and the limitations that this PR imposes. Briefly, these are: if you want to have use type class constraints on the parsing monad without turning on `-XNoMonomorphismRestriction`

  * You must have a version of GHC >= 7.10.3
  * Otherwise, you must add type signatures to all of your productions

### DIFF
--- a/src/ProduceCode.lhs
+++ b/src/ProduceCode.lhs
@@ -107,14 +107,14 @@ Produce the complete output file.
 >                             . str "#endif" . nl
 >                       | otherwise = id
 >
->    n_types = length (filter isNothing (elems nt_types))
+>    n_missing_types = length (filter isNothing (elems nt_types))
 >    happyAbsSyn = str "(HappyAbsSyn " . str wild_tyvars . str ")"
->      where wild_tyvars = unwords (replicate n_types "_")
+>      where wild_tyvars = unwords (replicate n_missing_types "_")
 >
 >    -- This decides how to include (if at all) a type signature
 >    -- See <https://github.com/simonmar/happy/issues/94>
 >    filterTypeSig :: (String -> String) -> String -> String
->    filterTypeSig content | n_types == 0 = content
+>    filterTypeSig content | n_missing_types == 0 = content
 >                          | otherwise = ifGeGhc710 content
 >
 >    top_opts =
--- a/src/ProduceCode.lhs
+++ b/src/ProduceCode.lhs
@@ -14,15 +14,15 @@ The code generator.
 >                                 interleave, interleave', maybestr,
 >                                 brack, brack' )
 
-> import Data.Maybe                     ( isJust, isNothing )
+> import Data.Maybe             ( isJust, isNothing, fromMaybe )
 > import Data.Char
 > import Data.List
 
-> import Control.Monad      ( forM_ )
+> import Control.Monad          ( forM_ )
 > import Control.Monad.ST
-> import Data.Bits          ( setBit )
-> import Data.Array.ST      ( STUArray )
-> import Data.Array.Unboxed ( UArray )
+> import Data.Bits              ( setBit )
+> import Data.Array.ST          ( STUArray )
+> import Data.Array.Unboxed     ( UArray )
 > import Data.Array.MArray
 > import Data.Array.IArray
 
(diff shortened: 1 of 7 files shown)


src/ProduceCode.lhs
```diff
@@ -90,16 +90,42 @@ Produce the complete output file.
...
+>
+>    -- If GHC is enabled, wrap the content in a CPP ifdef that includes the
+>    -- content and tests whether the GHC version is >= 7.10.3
+>    ifGeGhc710 :: (String -> String) -> String -> String
+>    ifGeGhc710 content | ghc = str "#if __GLASGOW_HASKELL__ >= 710" . nl
+>                             . content
+>                             . str "#endif" . nl
+>                       | otherwise = id
+>
+>    n_types = length (filter isNothing (elems nt_types))
```
simonmar: should be `n_missing_types` perhaps?

simonmar:
Does this fix the examples that were broken? (maybe only if you have GHC >= 7.10?)

harpocrates:
This fixes all regressions (note that CI checks also GHC 7.6 and 7.8)

  * all grammars that worked in Happy 1.19.5 still work
  * all grammars that depended on the feature in https://github.com/simonmar/happy/pull/49 and worked in 1.19.6 and 1.19.7 still work
  * some grammars that never could have worked before but have GHC >= 7.10.3 now also work

Does that clarify things?

harpocrates:
Friendly ping.

If there is anything in this PR that doesn't sit well with you, I don't mind adjusting it, explaining it, and/or closing it.

simonmar:
Thanks for the fix, and sorry for the delay.

PR #110, opened 2017-12-23
Add Paths_happy to other-modules

bgamari:
This silences the missing-home-module warning.

### DIFF
--- a/happy.cabal
+++ b/happy.cabal
@@ -154,6 +154,7 @@ executable happy
   extensions: CPP, MagicHash, FlexibleContexts
   ghc-options: -Wall
   other-modules:
+        Paths_happy
         AbsSyn
         First
         GenUtils


simonmar:
What warning? This seems dubious to me: `Paths_happy` is generated by Cabal itself, so why should we have to add it to `other-modules`?

asr:
> What warning? 

```
$ cabal install
...
Building executable 'happy' for happy-1.19.8..

<no location info>: warning: [-Wmissing-home-modules]
    These modules are needed for compilation but not listed in your .cabal file's other-modules: Paths_happy
```
Tested with:

```
$ ghc -V
The Glorious Glasgow Haskell Compilation System, version 8.2.2

$ cabal --version
cabal-install version 2.0.0.1
compiled using version 2.0.1.0 of the Cabal library
```

> so why should we have to add it to `other-modules`?

Because the `Paths_happy`module *explicitly* is used for Happy (grepping result):

```
./src/ProduceGLRCode.lhs:17:> import Paths_happy ( version )
./src/ProduceCode.lhs:9:> import Paths_happy            ( version )
./src/Main.lhs:12:> import Paths_happy
./src/Info.lhs:9:> import Paths_happy            ( version )
```
  
Note that if `Paths_happy` wasn't used, cabal would not generate the warning.

simonmar:
Well this still seems a bit silly to me, but ok.

PR #120, opened 2018-06-03
Remove an unnecessary unsafeCoerce#

treeowl:
For some reason, `Main.lhs` was using `unsafeCoerce#` to coerce
between a type and *itself*. Quit that.


simonmar:
Good catch, I guess it was required at some point in the distant past.

PR #128, opened 2018-08-12
bump for GHC 8.6.1

vmchale:
This bumps Cabal bounds to allow building with GHC 8.6.1 and removes a deprecated function from `Setup.lhs`.


erikd:
Since you're adding a new compiler version, its probably a good idea to add it to the `.travis.yaml` file as well.

hvr:
@erikd this looks like on of those transient network glitches which resolve when you simply retry the builds a bit later

PR #133, opened 2019-02-05
Generate 'let' for non-terminals

int-index:
The manual of `happy` [claims](https://www.haskell.org/happy/doc/html/sec-type-signatures.html) it supports polymorphic types:

> It is possible to have productions with polymorphic or overloaded types. However, because the type of each production becomes the argument type of a constructor in an algebraic datatype in the generated source file, compiling the generated file requires a compiler that supports local universal quantification. GHC (with the -fglasgow-exts option) and Hugs are known to support this.

In practice, this turned out not to be the case with the `-c` option.

----

Commit https://github.com/simonmar/happy/commit/3882dacfdb882da3bc06d83f22e659c16d27d754 replaced `let`-bindings of `happy_var_n` with  `case` to avoid lazy pattern matching.

However, GHC treats `let` and `case` quite differently when it comes to rank-N types. @RyanGlScott suggested it has to do with deep skolemization. I do not understand the issue fully, but here's a minimal example to demonstrate:

```
{-# OPTIONS -Wall #-}
{-# LANGUAGE RankNTypes #-}
{-# LANGUAGE ScopedTypeVariables #-}

module LetCase (f) where

f :: (forall b. b -> b) -> (forall b. b -> b)
f x =
   --let z = x in    (z :: forall b. b -> b)
   case x of z -> (z :: forall b. b -> b)
```
This program fails with

```
LetCase.hs:10:20: error:
    • Couldn't match type ‘b0’ with ‘b1’
        because type variable ‘b1’ would escape its scope
      This (rigid, skolem) type variable is bound by
        an expression type signature:
          forall b1. b1 -> b1
        at LetCase.hs:10:25-40
      Expected type: b1 -> b1
        Actual type: b0 -> b0
    • In the expression: (z :: forall b. b -> b)
      In a case alternative: z -> (z :: forall b. b -> b)
      In the expression: case x of { z -> (z :: forall b. b -> b) }
    • Relevant bindings include
        z :: b0 -> b0 (bound at LetCase.hs:10:14)
   |
10 |    case x of z -> (z :: forall b. b -> b)
   |  
```

Replacing `case` with the commented out `let` solves the issue and the code compiles.

However, we can't just go back to using `let`. There are two things to take into account:

1. We'd like to keep pattern matches strict, and preferably without depending on `-XBangPatterns`.
2. We want this to work even with `-GADTs` (and, consequently, `-XMonoLocalBinds`).

We can address these issues as follows:

1. Keep using `case` for terminals. For non-terminals, the match is always a variable (irrefutable) anyway, so we gain nothing by using `case`. Therefore: `let` for non-terminals, `case` for terminals.
2. Add an in-place type signature for each `let`-binding to sidestep `-XMonoLocalBinds`.

----

Here's an example of a grammar that used to fail before this patch (I also added it to the testsuite as `rank2.y`):

```
ib :: { (Int, Bool) }
   : f { ($1 5, $1 True) }

f :: { forall a. a -> a }
  : { id }
```

The `ib` production tried to instantiate `id :: forall a. a -> a` as `Int` for the first component and as `Bool` for the second component. GHC skolemized `a` leading to the following error message:

```
rank2.agc.hs:92:35: error:
    • Couldn't match expected type ‘Bool’ with actual type ‘Int’
    • In the expression: happy_var_1 True
      In the first argument of ‘happyIn4’, namely
        ‘((happy_var_1 5, happy_var_1 True))’
      In the expression: happyIn4 ((happy_var_1 5, happy_var_1 True))
   |
92 |                  ((happy_var_1 5, happy_var_1 True)
   |                                   ^^^^^^^^^^^^^^^^

rank2.agc.hs:92:47: error:
    • Couldn't match expected type ‘Int’ with actual type ‘Bool’
    • In the first argument of ‘happy_var_1’, namely ‘True’
      In the expression: happy_var_1 True
      In the first argument of ‘happyIn4’, namely
        ‘((happy_var_1 5, happy_var_1 True))’
   |
92 |                  ((happy_var_1 5, happy_var_1 True)
   |
```

Why? Let's take a look at the generated reduction rule:

```
happyReduction_1 happy_x_1
	 =  case happyOut5 happy_x_1 of { happy_var_1 -> 
	happyIn4
		 ((happy_var_1 5, happy_var_1 True)
	)}
```

We want `happy_var_1 :: forall a. a -> a` (to instantiate it as `Int` in one call and `Bool` in another), but doing a `case` leads to skolemization of `a`.

After this patch, the generated reduction rule uses `let` with a type signature:

```
happyReduce_1 = happySpecReduce_1  0# happyReduction_1
happyReduction_1 happy_x_1
	 =  (let {happy_var_1 :: forall a. a -> a; happy_var_1 = happyOut5 happy_x_1} in
	happyIn4
		 ((happy_var_1 5, happy_var_1 True)
	))
```

This compiles without error. The type signature is added only if the user has provided it.

----

As I said, I preserved `case` matches for terminals. Here's example output from the `monad002.ly` test case:

```
	 = happyThen ((case happyOutTok happy_x_2 of { (TokenVar happy_var_2) -> 
	(let {happy_var_4 :: Exp; happy_var_4 = happyOut4 happy_x_4} in
	(let {happy_var_6 :: Exp; happy_var_6 = happyOut4 happy_x_6} in
	( \s l -> ParseOk (Let l happy_var_2 happy_var_4 happy_var_6))))})
	) (\r -> happyReturn (happyIn4 r))
```

It uses `case` for `happy_var_2` (which is a terminal) and `let` for `happy_var_4` and `happy_var_6` (which are non-terminals). So, strictness is preserved where necessary.

----

PS. I also bumped the version to 1.19.10 and I would appreciate a release with this fix. Thanks!


int-index:
The CI failure seems unrelated to the patch:

```
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
W: GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: The repository 'https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease' is not signed.
N: Data from such a repository can't be authenticated and is therefore potentially dangerous to use.
N: See apt-secure(8) manpage for repository creation and user configuration details.
W: There is no public key available for the following key IDs:
6B05F25D762E3157  
E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/main/binary-amd64/Packages  Writing more data than expected (804385 > 804292)
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo apt-get update" failed and exited with 100 during .
Your build has been stopped.
```

I have verified that it works by running `cabal v1-test` locally.

int-index:
This solution is incomplete and I think I have a better one (which I will finalize and post tomorrow).

Issue #131, opened 2018-12-03
qualify Prelude functions in templates?


jvoigtlaender:
In a `.y` file, I would like to define my own version of the `div` function. So this is what I have there:
```haskell
import Prelude hiding (div)
import qualified Prelude

...
define and use my own div-function somewhere, possibly defined via Prelude.div in some way
... 
```
But this does not work, because the generated `.hs` file does use the original `Prelude` version of `div` without qualification, so actually my new version of `div` is chosen, which happens to have a different type.

The concrete code that triggers the conflict in my case appears to be from the lines
```haskell
readArrayBit arr bit =
    Bits.testBit IBOX(indexShortOffAddr arr (bit `div` 16)) (bit `mod` 16)
```
in `templates/GenericTemplate.hs`, but more generally the issue seems to be various uses of `Prelude` functions in Happy generated code.

Wouldn't it be better if the above and similar references were to `Prelude.div` instead of just `div`, and likewise for other `Prelude` functions?

simonmar:
Yes, it would be nice to fix this properly. Pull requests accepted!

jvoigtlaender:
Are the `.hs` files in the `templates` subdirectory all one would have to touch/look through?

Or are there other places, in `src`, that also generate code that has both of the following properties:
1. calls `Prelude` or other standard functions not also generated by Happy itself;
2. is joined together, in a single output module, with user code from an input `.y` file?

jvoigtlaender:
The point in 2. being: "*declarations* from an input `.y` file". Of course, all the expressions in productions' semantic actions are also user code, but these should be harmless for the issue at hand. The declarations of functions in a user's Haskell code blocks taken over from `.y` into output are the problem.

So places, in `src`, where semantic action code is taken over into the output module are not problematic. But if something in `src` generates its own calls to `div` or other `Prelude` functions, then those "something" places could be relevant to fix. I don't know how to search for such places. But anyway, I could focus on a fix for the files in `templates` first.

simonmar:
I'm fairly sure that you would also need to look at the parts of Happy that generate code, in particular https://github.com/simonmar/happy/blob/master/src/ProduceCode.lhs, and the part where we generate the imports, https://github.com/simonmar/happy/blob/master/src/Main.lhs#L490

But if you add `{-# LANGUAGE NoImplicitPrelude #-}` and `import qualified Prelude` and then fix everything enough to make the tests pass, that should be enough.

jvoigtlaender:
[...]

PR #135, opened 2019-04-02
Allow usage of qualified Prelude in templates

Kellador:
Should resolve [#131](https://github.com/simonmar/happy/issues/131);
allowing the usage of qualified Prelude in templates, as specified in the issue.

A set of all testcases modified to use NoImplicitPrelude and qualified Prelude is included in a seperate directory inside the tests directory.


simonmar:
Thanks for doing this! Two suggestions:

* Maybe it would be better for Happy to inject `import qualified Prelude as HappyPrelude` in the imports, and then qualify everything with `HappyPrelude`? That way we aren't relying on the user to import Prelude with a known name.

* I note that you copied all the tests. Rather than copying them in the repository, would it be possible to auto-generate the modified tests, perhaps using CPP or something? I'd rather not have multiple copies of the tests if possible, since they could easily diverge accidentally.

Kellador:
Thanks for the suggestions, 
injecting the import definitely sounds like a good idea, and I will look into how to best auto-generate the modified tests as well.

Kellador:
`import qualified Prelude as HappyPrelude` is now injected and everything qualified with `Prelude` before is now qualified with `HappyPrelude` instead.
This alone would however make all old grammars incompatible, since they would then have to import Prelude themselves without qualification, so I've added a new cli option: `--noprelude`
https://github.com/Kellador/happy/blob/master/src/Main.lhs#L439-L440
Without this flag the default behaviour is injection of both:
`import qualified Prelude as HappyPrelude` and `import Prelude`,
whereas with the flag, the latter import is omitted; 
A side effect of this is that with the flag it is not necessary to declare `{-# LANGUAGE NoImplicitPrelude #-}`.

Both tests for `HappyPrelude` usage with the `--noprelude` flag and basic `Prelude` are now generated via the C preprocessor.
For this all relevant tests are preprended with 
`#ifndef QUALIFIEDPRELUDE
#define QUALIFIEDPRELUDE Prelude
#endif`
and new targets were added to the tests Makefile (https://github.com/Kellador/happy/blob/master/tests/Makefile#L40-L126).

This in effect doubles up all tests, which might be why the Travis CI build is currently timing out.

simonmar:
> import qualified Prelude as HappyPrelude is now injected and everything qualified with Prelude before is now qualified with HappyPrelude instead.
This alone would however make all old grammars incompatible, since they would then have to import Prelude themselves without qualification

I don't follow that - why wouldn't it work to just inject `import qualified Preliude as HappyPrelude` and then refer to all Prelude identifiers qualified by `HappyPrelude.`? We don't have to add `{-# LANGUAGE NoImplicitPrelude #-}` too, we can let the user choose whether to do that or not.

Kellador:
> I don't follow that - why wouldn't it work to just inject `import qualified Preliude as HappyPrelude` and then refer to all Prelude identifiers qualified by `HappyPrelude.`? We don't have to add `{-# LANGUAGE NoImplicitPrelude #-}` too, we can let the user choose whether to do that or not.

Because once Prelude is imported under a qualified name it is not implicitly imported anymore;

If the extra unqualified `import Prelude` is omitted all tests without the HappyPrelude qualifiers fail, even without `{-# LANGUAGE NoImplicitPrelude #-}`.

simonmar:
Oh, I'd forgotten about that. How unfortunate.

Kellador:
Giving Travis CI a little more time before timing out did the trick, so I pushed that along with the last few fixes to make it work for the older GHC versions.

Not sure about appveyor, as that looks to have been broken for a while now and fixing that would not be related to this pull request.

jvoigtlaender:
Thanks @Kellador! The work done here was essential for getting https://github.com/simonmar/happy/pull/156 right.

PR #136, opened 2019-04-12
Fix for GHC 8.8.1

vmchale:
I'm not 100% sure what GHC 8.8.1 will look like when released, but in any case: this fixes it so that it works with the GHC 8.8.1 in hvr's ppa.


vmchale:
Looks like the CI failure is something upstream :(

simonmar:
Thanks.

PR #142, opened 2019-06-16
Setup.hs: Compatibility with Cabal-3

bgamari:
https://github.com/haskell/cabal/pull/3693 deprecated `rawSystemProgramConf` in favor of `runDbProgram`.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -44,7 +44,7 @@ symbols cs = case lex cs of
               _ -> []
 
 myPostBuild _ flags _ lbi = do
-  let runProgram p = rawSystemProgramConf (fromFlagOrDefault normal (buildVerbosity flags))
+  let runProgram p = runDbProgram (fromFlagOrDefault normal (buildVerbosity flags))
                                           p
                                           (withPrograms lbi)
       cpp_template src dst opts = do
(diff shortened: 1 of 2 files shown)


RyanGlScott:
For backwards compatibility with older versions of GHC (which bundle `Cabal`s that don't support `MIN_VERSION` macros in `Setup` scripts), I think you'll need to add the following lines to the top of `Setup.hs`:

```hs
#ifndef MIN_VERSION_Cabal
#define MIN_VERSION_Cabal(x,y,z) 0
#endif
```

RyanGlScott:
Better yet, you don't even need to use CPP at all. The `runDbProgram` function has existed for long enough that it can just be used in place of `rawSystemProgramConf`. The oldest version of GHC that `happy` currently supports is 7.6.3 which bundles `Cabal-1.16`, and even that release featured `runDbProgram` (see [here](http://hackage.haskell.org/package/Cabal-1.16.0.3/docs/Distribution-Simple-Program.html#v:runDbProgram)).

shayne-fletcher-da:
Great news @bgamari @RyanGlScott @simonmar and thanks ! Do we have an ETA for a release to see this out the door?

PR #150, opened 2019-09-30
Replace type synonyms with proper types, add docs

harpocrates:
This commit changes no new functionality, but hopefully it makes the
code more readable and less prone to accidental type confusion. All
refactoring is very mechanical:

  * replace type synonyms of large tuples with proper types
  * add conservative comments to data types and functions
  * add a handful of type signatures

### DIFF
--- a/src/Grammar.lhs
+++ b/src/Grammar.lhs
@@ -9,7 +9,7 @@ Here is our mid-section datatype
 > module Grammar (
 >       Name,
 >
->       Production, Grammar(..), mangler, ErrorHandlerType(..),
+>       Production(..), Grammar(..), mangler, ErrorHandlerType(..),
 >
 >       LRAction(..), ActionTable, Goto(..), GotoTable, Priority(..),
 >       Assoc(..),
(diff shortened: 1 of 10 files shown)


src/Grammar.lhs
```diff
@@ -34,7 +34,14 @@ Here is our mid-section datatype
 
 > type Name = Int
 
-> type Production = (Name,[Name],(String,[Int]),Priority)
+> data Production
+>       = Production Name [Name] (String,[Int]) Priority
+
+#ifdef DEBUG
+
+>       deriving Show
```
Ericson2314: Why not always derive?

Ericson2314: Ahh `Name`. I thought this was all Prelude stuff stuff in there, but I was wrong. Nevermind, :+1:.

int-index:
The AppVeyor CI failure is unrelated to the patch and is fixed in #154

simonmar:
Thankyou! This code is so old :)

PR #152, opened 2019-10-03
Remove deprecated flag small_base

mary-avocado:
This patch removes an unused configuration flag. It has been deprecated for 5 years.


int-index:
The AppVeyor CI failure is unrelated to the patch and is fixed in #154

PR #170, opened 2020-08-15
No bootstrapping

### LINKED ISSUE #169: No bootstrapping
Currently, building `happy` requires a pre-built binary of `happy`.

It is, surely, elegant, in the same way a self-hosting compiler is elegant. But it also makes building purely from source more complicated than needed.

What we could do instead is parse `.y` and `.ly` files using parser combinators. But it would also be nice to avoid a dependency on another parsing library. Also, the currently existing hand-rolled lexers are written in continuation-passing style, so we'd need a library compatible with that.

So I developed a small, bespoke parsing library, specifically for `happy`, and it seems to work nicely. Patch here: #170

int-index:
Before this patch, building `happy` required a pre-built binary of `happy`. This was elegant in the same way a self-hosting compiler is elegant. But it also made building purely from source more complicated than needed.

This patch introduces a small, bespoke parsing library, and applies it for parsing `.y` and `.ly` files.

Now `happy` doesn't depend on itself, and can be built using just GHC.

Closes #169


int-index:
We discussed this with @simonmar elsewhere and concluded that a better solution (for now) would be to check in Happy-generated files.

Ericson2314:
Could we do it both ways, using compiler flags or something to switch between the too? Beyond bootstrapping, I think this is useful to test happy.

Ericson2314:
@simonmar Maybe see:

1. @bgamari worrying about making happy a submodule in GHC: https://mail.haskell.org/pipermail/ghc-devs/2020-August/019111.html
2. Me responding suggesting something along these lines https://mail.haskell.org/pipermail/ghc-devs/2020-December/019397.html

It is nice to have the submodule so we can have GHC branches using alex/happy branches, whether to demonstrate new features, or as in the recent case, that a branch changing the primops works at all.

Issue #167, opened 2020-05-11
Consider a recusive ascent/descent backend


sgraf812:
See https://github.com/djspiewak/scala-bison, its [tool paper](https://www.sciencedirect.com/science/article/pii/S1571066110001118) and [the original paper on recursive ascent/descent](https://link.springer.com/chapter/10.1007/3-540-53669-8_70).

sgraf812:
FWIW, I now have a student tackling this for his bachelor's thesis. So eventually there will be at least one PR.

sgraf812:
(And half a year later, there's #174 🎉 )

andreasabel:
@sgraf812 @simonmar 
Basic question: why should this go into `happy`?  Can't you make a separate parser generator that is compatible with `happy`, i.e., reads `.y` files?

Ideally, of course, it is great to have added functionality.

From a pragmatical side, there are down sides:
- The new component is an additional maintenance burden.
- Your student is not likely to commit to maintaining this for at least the next 5-10 years.
- `happy` is core to the Haskell community as a whole and needs to be __very stable__.

Note that we already have a broken backend in `happy`: `--glr` with its original contributors vanishing after completing their thesis work.

I think these aspects should be taken into account when adding functionality to `happy`, in particular when it is substantial.

Sorry for playing the devil's advocate here.

My reservation comes from experiences with maintaining the [BNFC](https://github.com/BNFC/bnfc) front-end to parser generators.  This tool has suffered from feature bloat from eager contributors who then moved on to other projects, already after 5 years not feeling responsible to maintain their components.  In the end, I had to weed out these components.  That is undesirable, because you never know whether some user out there depends on these features.  

I think all of `happy` should be in good shape before we accept new components.
And new components should come with a signed maintenance pledge for _at least_ 5 years.

sgraf812:
Those are very good points. I agree, `happy` is not actually in a shape that accommodates any more extension. But what I like about #174 is that it is a drop-in replacement for `happy`.

In an ideal world, it would make sense to split `happy` into three components:

1. `happy-frontend`: Parsing y-Files (including GHC-specific stuff), return an AST representing the grammar+entrypoints
2. `happy-middleend`: grammar analysis that generates LALR(1) states, producing a (potentially ambiguous) `ActionTable` and `GotoTable`
3. `happy-codegen`: Different codegen strategies, like table-based LALR and GLR.

[...]

PR #174, opened 2020-10-31
Implement a typed recursive ascent-descent backend

knothed:
This PR extends `happy` by a _Recursive Ascent-Descent_ backend and therefore implements #167.

In particular, the following options are added:

* `--cb-lalr`: use `happy`'s LALR states to create code in well-typed, continuation-based style. This style is due to Hinze and Paterson as described in [Derivation of a Typed Functional LR Parser](https://www.cs.ox.ac.uk/ralf.hinze/publications/TypedLR.pdf).

* `--cb-rad`: produce a **recursive ascent-descent** (RAD) parser, also using well-typed, continuation-based style. The RAD states are generated based on `happy`'s LALR states.
* `--cb-rad --optims`: optimize the generated code for speed by employing some optimizations.
* `--types`, `--comments`: annotate all functions with their explicit type (if possible) / with a comment. Sometimes, explicit types are required in order for GHC to build the generated parser.

In the following, I'll summarize the intention behind using recursive ascent-descent parsing, how it interacts with `happy`, and what advantages it brings.

**TL;DR:** 

- RAD (or GLC) parsing takes advantage of redundancies in LALR states to reduce the number of states, and the number of actions per state, significantly.
- In combination with the directly-executable, continuation-based form, the parser is faster than a table-based variant.

* Replacing the current, `happy`-generated, Haskell parser of the Glasgow Haskell Compiler with a recursive ascent-descent variant speeds up parsing of Haskell files by GHC by around 10%.

  In smaller grammars (e.g. JSON), recursive ascent-descent parsers beat `happy`-generated ones by a factor of 3x to 4x.

The remainder is structured as follows:

1. What is Recursive Ascent-Descent?
2. Implementation
3. Compatibility with existing `happy`-features
4. Speed and size comparison
5. TBD

---

# What is Recursive Ascent-Descent?

A recursive descent parser contains one parsing routine per nonterminal. This routine decides, based on the lookahead token, which production to follow. Then, all symbols of this production's right-hand side are parsed consecutively.

In contrast, LALR parsers can pursue multiple productions at once. An LALR state can have multiple core items, all of which describe a possible position inside a production in which the parser could currently be. 

LALR parsers are very often found in table-based form. There also exists _recursive ascent_ – similar to recursive descent, a recursive ascent parser has one function per state. Shift and goto actions are executed by calling a different (or the same) state function, while reduce actions pop the function call stack after calling a semantic action.

Because of its direct, function-based nature, a recursive ascent parser is often faster than a respective table-based parser. This comes with a burden: The compiled code is much larger. As an example, the table-based `happy`-generated LALR parser of GHC is 2.5 MB. A recursive-ascent form is a little faster, but is 33 MB large.

Now comes recursive ascent-descent: By merging and reusing similarities of multiple LALR states, the remaining states can be made smaller or partly unnecessary. This works by splitting each production at its _recognition point_: Up to the recognition point, parsing proceeds bottom-up using the RAD states, similar to LALR parsing. Once the recognition point of a rule is reached, it can be decided – based on the lookahead token – whether the rule is active or not. Remember: in an SLL grammar, this can be decided at the very beginning of each rule. After deciding that a rule is active, its remaining symbols are parsed consecutively, in a top-down fashion.

Bottom-up is only used while required, and the switch to top-down happens as early as possible.

Consider a nonterminal `Value` which appears in the right-hand side of multiple rules. Every time it appears _after_ the recognition point, this `Value` parse proceeds identically. In an LALR parser, each of these parsing situations would require separate state sequences, whereas a RAD parser extracts them into a single procedure – `parseValue`. This procedure begins by entering into an _entry state_ – the entry state of `Value`. Then, bottom-up parsing proceeds until the `Value` has fully been parsed and control is passed back to the calling rule procedure.

What is this good for? Because of the great savings of states and actions, a recursive ascent-descent parser is much smaller than a respective recursive ascent parser. Take another look at GHC's Haskell parser:

|                  | table-based (happy-generated) | recursive ascent (LALR) | recursive ascent-descent (RAD) |
| ---------------- | ----------------------------- | ----------------------- | ------------------------------ |
| Size of Parser.o | 2.4 MB                        | 33 MB                   | 7.6 MB                         |

The RAD parser is, of course, still larger than the status-quo, table-based, happy-generated parser, but not by a factor of 14 but only of 3. This goes along with a speedup of around 10%, which comes through the continuation-based form that we have chosen for code generation.

For more detailed explanations of what RAD is, how it works, how RAD states are generated from LALR states, how recognition points are found, how continuation-based code is generated and used in conjunction with happy, and for performance comparisons with LALR parsers, I refer to my bachelor's thesis: [A Typed Recursive Ascent-Descent Backend for Happy](https://pp.ipd.kit.edu/uploads/publikationen/knothe20bachelorarbeit.pdf).

Originally, recursive ascent-descent parsing was described by [Horspool](http://boost-spirit.com/dl_docs/rad.pdf) and builds upon [Generalized Left Corner parsing](https://dl.acm.org/doi/10.1145/512950.512966).

---

# Implementation

We shortly talk about how we implemented the new options.

- `--cb-lalr`: uses `happy`'s LALR states to create code in well-typed, continuation-based style. This code is often faster than code generated by happy, but much larger.
  1. `happy` is executed normally, until conflict reporting and info file creation.
  2. Then, we branch off into `RADCodeGen_LALR.hs` to generate the code.
- `--cb-rad`: creates a RAD parser in well-typed, continuation-based style. This code is often faster than code generated by happy, and not much larger.
  1. `happy` is executed normally, until conflict reporting and info file creation.
  2. `happy`'s LALR states are used to determine the recognition points of all productions.
  3. RAD states are created based on `happy`'s LALR states (in `RADStateGen.hs`). This process respects features like `happy`'s conflict resolution and the error token. It is described in detail [here](https://pp.ipd.kit.edu/uploads/publikationen/knothe20bachelorarbeit.pdf#page=21).
  4. Finally, code is created similar as in the `--cb-lalr` case (`RADCodeGen.hs`). Code generation is described in detail [here](https://pp.ipd.kit.edu/uploads/publikationen/knothe20bachelorarbeit.pdf#page=43).
- `--cb-rad --optims`: we employ further optimizations to further speed up the generated parser. Currently, these include rule function inlining and eta-expansion of continuation functions.

---

# Compatibility with existing `happy`-features

Our backend supports most of `happy`'s features and directives. It supports all options that are used in GHC's Haskell grammar and is therefore powerful enough to produce a Haskell parser.

Especially, because `happy`'s LALR states (i.e. action and goto tables) are used for RAD state generation, conflict resolution happens just as before, and directives like `%left` or `%expect` work as desired.

Monadic parsers and lexers are supported – the generated code is adapted to the monadic style. Partial parsers are also supported, just like the error token.

Features that are **not** (yet) supported are:

- Using `%monad` _without_ `%lexer`, or vice versa. It doesn't seem anyone is doing this, therefore we didn't really care about this case.
- `{%^ ... }` and `{%% ... }` semantic actions. If these are desired, they could be implemented. Of course, normal monadic actions (`{% ... }`) are supported.
- Attribute grammars.
- GLR parsing. This is, of course, incompatible with RAD parsing.
- `happy`-specific options like `-a`, `-c`, and `-g` have no equivalent in continuation-based code. We are however currently working on a `--strict` implementation.

---

# Speed And Size Comparison

The detailed results can be seen [here](https://pp.ipd.kit.edu/uploads/publikationen/knothe20bachelorarbeit.pdf#page=61). We just briefly look at two results now: small grammars, and GHC.

First we look at two small grammars – an expression grammar and a JSON grammar. We compare the result of the fastest `happy`-generated parser (which could be `happy`, `happy -acg` or `happy -acg --strict`) with our `--cb-rad` and `--cb-rad --optims` parsers. The following table shows the average plain parsing times (using a large input file), without lexing.

| Grammar    | fastest `happy` | `--cb-rad` | `--cb-rad --optims` |
| ---------- | --------------- | ---------- | ------------------- |
| Expression | 90.0 ms         | 35.7 ms    | 28.0 ms             |
| JSON       | 61.1 ms         | 16.1 ms    | 15.3 ms             |

Our parser beats the `happy`-generated ones by a factor of 3-4. 

Now we consider GHC. We built GHC 8.10.1 with its normal `happy`-generated parser, and with a Parser.hs in RAD-form. Then we used `-ddump-timings` to obtain the parsing times when parsing Haskell files. We parsed four large files. Each time we compared both the stage-1 and the stage-2 parser.

| File | Stage | status quo (`happy -acg --strict`) | `--cb-rad --optims` | Speedup |
| ---- | ----- | ---------------------------------- | ------------------- | ------- |
| 1    | 1     | 3780 ms                            | 3051 ms             | 19.4%   |
| 1    | 2     | 3459 ms                            | 3162 ms             | 8.6%    |
| 2    | 1     | 60.1 ms                            | 55.5 ms             | 7.7 %   |
| 2    | 2     | 47.0 ms                            | 45.8 ms             | 2.5%    |
| 3    | 1     | 432 ms                             | 387 ms              | 11.4%   |
| 3    | 2     | 404 ms                             | 382 ms              | 5.5%    |
| 4    | 1     | 784 ms                             | 680 ms              | 13.3%   |
| 4    | 2     | 748 ms                             | 713 ms              | 4.9%    |

We believe that we can expect even higher speed-ups by employing further optimizations. Especially, there are some performance regressions from stage-1 to stage-2 using our RAD parsers (file 1 and file 4) that _do not_ happen for the `happy`-generated parsers. Getting rid of these regressions would mean higher speedups.

Finally here is the above table again comparing the Parser.o sizes:

|                  | status quo | `--cb-lalr` | `--cb-rad --optims` |
| ---------------- | ---------- | ----------- | ------------------- |
| Size of Parser.o | 2.4 MB     | 33 MB       | 7.6 MB              |

---

# TBD

Some time, before or after merging this PR, some things must still be done:

- [ ] Update README.
- [ ] Update documentation; maybe explain RAD parsing (and continuation-based parsing) in its own chapter.
- [ ] Add examples and test cases.
- [ ] Generate a RAD-specific info file on `-i`. Currently, no RAD-specific information is given to the user (except when they use the `--comments` flag and look at the generated code).
- [ ] Further size and performance optimizations, for example:
  - [ ] Implement `--strict` flag
  - [ ] Use short-cut fusion to optimize away `repeatTok/lexerWrapper` call pairs
- [ ] Fix performance regressions from stage-1 to stage-2 GHC parsers.
- [ ] Try reducing the compilation time of generated parsers. They can currently take a long time to compile.


src/Main.lhs
```diff
@@ -202,6 +206,84 @@ Add any special options or imports required by the parsing machinery.
...
+>       in
+
+>       if OptCB_LALR `elem` cli
+>       then
+>       let (isMonad, _, parserType, _, _) = monad g
+>       
+>           ptype = case (Grammar.lexer g, isMonad) of
+>             (Nothing, False) -> LALR.Normal
+>             (Nothing, True) -> LALR.Monad
+>             (Just _, False) -> error "%lexer without %monad not supported in RAD"
```
sgraf812, line 227: We should think about whether and how we can support it.

src/Main.lhs
```diff
@@ -202,6 +206,84 @@ Add any special options or imports required by the parsing machinery.
...
+>             (Just _, True) -> LALR.MonadLexer
+>         
+>           options = LALR.GenOptions {
+>             LALR.ptype = ptype,
+>             LALR.wrapperType = if parserType == "Parser" then "HappyP" else "Parser",
```
sgraf812, line 232: Well, this can't be right, can it? The info you seek is encoded in the `Grammar`'s `monad` field.

src/RADCodeGen.hs
```diff
@@ -0,0 +1,557 @@
...
+  import GHC.Arr ((!), indices)
+  
+
+  data ParserType = Normal | Monad | MonadLexer deriving (Eq, Show)
+  
+  data GenOptions = GenOptions {
+    ptype :: ParserType,
+    
+    wrapperType :: String, -- e.g. "Parser"
+    errorTokenType :: String, -- e.g. "ErrorToken"
```
sgraf812, line 20: Is this something that needs configuring? What do the other Backends do?

src/RADCodeGen.hs
```diff
@@ -0,0 +1,557 @@
...
+  import Data.Text (pack, unpack, replace)
+  import GHC.Arr ((!), indices)
+  
+
+  data ParserType = Normal | Monad | MonadLexer deriving (Eq, Show)
+  
+  data GenOptions = GenOptions {
+    ptype :: ParserType,
+    
+    wrapperType :: String, -- e.g. "Parser"
```
sgraf812, line 19: This is hardly enough, I think. See my comments above

src/RADCodeGen.hs
```diff
@@ -0,0 +1,557 @@
...
+  import RADTools (XGrammar(..), showItem, showProd, lhs, showRecognitionPoint, recognitionPoints, rhsAfterDot)
+  import RADStateGen
+  import Control.Monad
+  import Data.List
+  import Data.Maybe
+  import Data.Text (pack, unpack, replace)
+  import GHC.Arr ((!), indices)
+  
+
+  data ParserType = Normal | Monad | MonadLexer deriving (Eq, Show)
```
sgraf812, line 14: You probably need more modes than that

src/RADCodeGen.hs
```diff
@@ -0,0 +1,557 @@
...
+  
+  replaceDollar a char = maybe a ($ char) (mapDollarDollar a)
+  
+  -------------------- GENCODE --------------------
+  -- Generate the full code
+  genCode :: GenOptions -> XGrammar -> [RADState] -> ActionTable -> GotoTable -> IO String
+  genCode opts x states action goto = do
+    return $ newlines 3 [languageFeatures, header', entryPoints', definitions', rules', parseNTs', parseTerminals', states', actions', footer'] where
+      languageFeatures
+        | rank2Types opts = newline $ map extension ["RankNTypes", "ScopedTypeVariables"]
```
sgraf812, line 52: I think that the user should activate `-XRankNTypes` if needed. Although I'm not so sure about `ScopedTypeVariables`...

Maybe you should grep for a line starting with `{-#` and cointaining `RankNTypes` to insert `ScopedTypeVariables` and to activate the annotation of type signatures in general.

sgraf812 (changes requested):
A bit of review feedback to get the discussion rolling. Generally quite nice, but we need to try harder to blend in with and reuse code of the existing implementation.

happy.cabal
```diff
@@ -161,12 +161,14 @@ executable happy
   build-depends: base < 5,
                  array,
                  containers >= 0.4.2,
+                 dom-lt >= 0.2.2,
```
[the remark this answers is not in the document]
phadej: It is some kind of a deal

- https://matrix.hackage.haskell.org/#/package/happy
- https://matrix.hackage.haskell.org/#/package/dom-lt

Currently `happy` is buildable with virtually any GHC you can get your hand on. With `dom-lt` it won't be true anymore.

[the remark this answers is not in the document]
sgraf812: Hmm. I wonder if we could remove the lower bound. Then we'd be fine, I think. AFAICT, the lower bound is because of https://github.com/AndreasPK/dom-lt/issues/2, but if 0.2.0 and 0.2.1 were marked broken, the solver wouldn't pick these.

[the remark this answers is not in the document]
phadej: That issue says that `dom-lt-0.1.3` is broken as well, so that won't work.

It might be easier to say that `happy-next` would work only with GHC-8.0+. I think it is fine, not small issue, but not huge either.

src/Main.lhs
knothed: No, this wrapperType is a wrapper _around_ the user-supplied monad type (see 4.2.1 in [here](https://pp.ipd.kit.edu/uploads/publikationen/knothe20bachelorarbeit.pdf): the user-defined monad is called `P`, and we define a wrapper and call it `Parser`). This wrapper is always called "Parser", except when the `Grammar`'s `monad` field is also called "Parser"; then we call the wrapper type "HappyP".

src/RADCodeGen.hs
knothed: No, this doesn't really need to be configured, we just need to choose some string for the dummy type. We could also fully remove the ErrorToken dummy type; this would change the number of arguments of affected continuations and semantic actions (so we would have to be more careful at other places), but should still work.

src/RADCodeGen.hs
knothed: The `ParserType` enumeration only reflects the operating modes introduced via the presence of `%monad` and `%lexer` directives. These affect whether wrapper types are required (because of lexerWrapper / repeatTok etc.) and some other places of code generation.

Therefore, we would, at most, need four of these values. Currently, only `Normal` and `MonadLexer` is really supported by my implementation, and the other options (`Monad` and possible `LexerOnly`) still need to be implemented. I could add this on the TBD list if requested. But I haven't found any happy grammars which were using either of `%monad` or `%lexer` without also using the other. Therefore I decided to, at the moment, just support `Normal` and `MonadLexer`.

src/Main.lhs
```diff
@@ -436,7 +526,21 @@ The command line arguments.
...
+>    Option [] ["optims"] (NoArg OptCB_RAD_Optimizations)
+>       "add optimizations such as eta-expansions and explicit rule-inlining to a RAD parser",
+>    Option [] ["cb-lalr"] (NoArg OptCB_LALR)
+>       "create a continuation-based LALR parser. Not compatible with most other options",
+>    Option [] ["types"] (NoArg OptCB_ShowTypes)
+>       "show function types for continuation-based parsers",
+>    Option [] ["comments"] (NoArg OptCB_ShowComments)
+>       "show comments for continuation-based parsers",
+>    Option [] ["forall"] (OptArg OptCB_ForallMatch "MATCH")
+>       "a string which is used to detect and handle higher-rank function types"
```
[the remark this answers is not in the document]
knothed: I thought the same, but what about a grammar which has a semantic type like `ForallStatement`? Or different variations of the word `forall`? What should be the string that is used to match for an actual `forall`-expression?

src/Main.lhs
```diff
@@ -436,7 +526,21 @@ The command line arguments.
...
+>    Option [] ["cb-rad-tuple"] (NoArg OptCB_RAD_TupleBased)
+>       "same as cb-rad, but uses tuples instead of continuations inside rule functions",
+>    Option [] ["optims"] (NoArg OptCB_RAD_Optimizations)
+>       "add optimizations such as eta-expansions and explicit rule-inlining to a RAD parser",
+>    Option [] ["cb-lalr"] (NoArg OptCB_LALR)
+>       "create a continuation-based LALR parser. Not compatible with most other options",
+>    Option [] ["types"] (NoArg OptCB_ShowTypes)
+>       "show function types for continuation-based parsers",
+>    Option [] ["comments"] (NoArg OptCB_ShowComments)
+>       "show comments for continuation-based parsers",
```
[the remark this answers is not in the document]
knothed: Yes, makes sense. The RAD-specific info file generation is a point on the todo list.

Actually, as the comments do not increase compiled code size, I would simply put them always in the generated code. There is no benefit in not adding comments.

happy.cabal
```diff
@@ -186,6 +188,11 @@ executable happy
         AttrGrammarParser
         ParamRules
         PrettyGrammar
+        RADCodeGen
+        RADCodeGen_LALR
```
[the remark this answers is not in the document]
knothed: True. There seems to be no advantage of the recursive ascent backend over the recursive ascent descent backend.

happy.cabal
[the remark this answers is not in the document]
knothed: Why exactly is `dom-lt-0.2.2` not compatible with GHC-7? Maybe this can be dealt with in `dom-lt`?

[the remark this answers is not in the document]
AndreasPK: > Why exactly is dom-lt-0.2.2 not compatible with GHC-7

Maybe the code would work. I honestly just haven't tried it with GHC 7 as I haven't used that old an GHC in years by now.

If someone put's the effort in to make it work/test it with GHC 7 I have no issue lowering the bounds.

PR #176, opened 2020-12-05
Get rid of uneeded windows-only imports

Ericson2314:


int-index:
I don't have a Windows machine to test it, but the CI has passed, so I will assume these imports were indeed unneeded

Issue #149, opened 2019-09-29
Happy and template haskell


Ericson2314:
I hope via https://github.com/ghc-proposals/ghc-proposals/pull/243 to make GHC able to use TH, but as GHC is not the sole user of happy, I figured I can get the ball rolling on brainstorming ideas now.

The first change I think would be to allow separating the grammar from the user written code. Happy could turn a `.y` files *without* any Haskell in between the braces into its representation of the grammar. Then, in another module, one writes all the production eliminators, named according to some convention. Finally, one imports happy as a library along with that grammar, and has a splice to generate what comes out of `.y` today.

Now, while the separation of grammar and eliminator seems like classic nice separating concerns, I'll wholly admit that using multiple completely separate ASTs with the same grammar is not likely to be useful in practice---it would be better to write a polymorphic AST and single set of polymorphic production eliminators.

The real benefit is now we have a chance to programmatically modify the grammar before parser codegen. Things like parameterized/macro rules and inlining (#148) could perhaps be done without as much, or any, support from Happy itself. User manipulation of the grammar could also allow it to be nicely split between multiple files, or more broadly, be compositionally constructed.

CC @sgraf812

int-index:
I have a [prototype of an EDSL](
https://github.com/int-index/euphoric/blob/fa75eb806f15f63a3f7a724b87065e036129d455/src/Euphoric.hs#L149-L169) to generate Happy grammars:

```
data ExToken =
  ExTokTrue |
  ExTokFalse |
  ExTokLPar |
  ExTokRPar

data ExRepr = ExPar ExRepr | ExBLit Bool

grammar :: Grammar ExToken (State Int) _
grammar = MkGrammar
  $ tm @"true"  [p|ExTokTrue|]
  $ tm @"false" [p|ExTokFalse|]
  $ tm @"(" [p|ExTokLPar|]
  $ tm @")" [p|ExTokRPar|]
  $ nt @"lit" @ExRepr
    do rule  @'["true"]            [|| \_ -> ExBLit True ||]
       rule  @'["false"]           [|| \_ -> ExBLit False ||]
  $ nt @"expr" @ExRepr
    do ruleM @'["(", "expr", ")"]  [|| \_ e _ -> ExPar e <$ modify (+1) ||]
       rule  @'["lit"]             [|| id ||]
  $ EndOfGrammar
```

With visible `forall` in terms, it would look become less noisy:

```
data ExToken =
  ExTokTrue |
  ExTokFalse |
[...]

PR #179, opened 2020-12-25
Get rid of template preprocessing

Ericson2314:
Instead of preprocessing an outer layer of CPP when building happy, just always produce code that uses CPP. Combined with #175, this means happy now has a perfectly bog standard build system, with Makefiles and extra steps strictly optional.

I gather Hugs, and possibly other Haskell implementations, out of the box doesn't support CPP, but I don't want this to stop us. Those can just manually run CPP on the generated code first.

Contains #175 and #178


int-index:
I was actually hoping we'd move in the opposite direction w.r.t. CPP, so that the output of `happy` is proper Haskell, not Haskell+CPP.

Otherwise I don't see how we could achieve #149

Ericson2314:
@int-index Well in the short term, I think it's really import that Happy and Alex have "bog standard" build systems so that we can make them submodules in GHC and never have to do things like https://hackage.haskell.org/package/alex-3.2.6/changelog that lead to https://gitlab.haskell.org/ghc/ghc/-/issues/19099 again. And the submodule would allow us to iterate on happy and TH in tandem for #149 too.

Now, I thought that was an impediment for #149 too, but I think I have a pretty fun solution for this.
Suppose we follow TTG and instead of having just:
```haskell
data Dec = ...
```
we have
```haskell
data Dec' a = ...
          | XDec a
type Dec = Dec' NoExtCon
```
Now we can do
```haskell
[d|
   foo = blah
   bar = blah
   $(XDec CppIf "Condition" [d| ...] CppEndif)
|] :: Dec' Cpp
```
CPP + TH without polluting TH itself with such a toxin!

simonmar:
Seems like a good idea.

PR #181, opened 2020-12-25
Remove sandboxing reminants from test makefile

Ericson2314:
Cabal sandboxing is long gone


tests/Makefile
```diff
@@ -92,13 +92,14 @@ HC_OPTS += -fforce-recomp
 check.%.y : %.y
 	@echo "--> Checking $<..."
 	$(HAPPY) $(TEST_HAPPY_OPTS) $< 1>$*.run.stdout 2>$*.run.stderr || true
+	sed -i '/^Up to date$$/d' $*.run.stdout $*.run.stderr
```
Ericson2314, line 95: This is so I can run the makefile locally with parallelism

Issue #187, opened 2021-03-19
Modernizing the Packaging 


Ericson2314:
We have some discussion in https://github.com/simonmar/happy/issues/167 and https://github.com/simonmar/happy/issues/149, but let's lay out the first steps in a separate issue.

 - [x] get rid of the extra template preprocessing step (https://github.com/simonmar/happy/pull/179)
 - [x] make bootstrapping optional (https://github.com/simonmar/happy/pull/175)
 - [ ] Make parser combinator bootstrap path simpler by reusing `ReadP` from base (see https://github.com/simonmar/happy/pull/175#issuecomment-752951456)

At this point the bootstrap will be as simple as it's ever been, and we still support GHCs 7 and up, so I would like to do a release.

Afterwords I would like to take more radical changes:

 - [x] Start splitting up package into (sub) libraries
     - [ ] Maybe factor out "run time" into one such library, rather than shoving it in the same file as the generated output from templates
 - [x] Maybe drop support for old cabal

sgraf812:
> Start splitting up package into (sub) libraries

I think @knothed already made some progress in that direction. Since hackage doesn't support multi public library packages yet, we settled on one cabal file for each subcomponent, which means more administrative overhead each release. Maybe some script/configure/Makefile can automate the process.

sgraf812:
The current example is at https://github.com/knothed/modular-happy. Still a few refactorings away from being feasible to merge back into happy mainline, but hopefully not too long. The general ideas of https://github.com/simonmar/happy/issues/167#issuecomment-780591344 are still valid, there's but one more package: `happy-executable` (maybe `happy-cli` would be better), that provides CLI parsing facilities. We were prveiously wrangling the different design options and arrived at a satisfying one that David is going to implement in the next weeks. Here's a sneak peek of a `main` function that includes support for the RAD extension (which ultimately will live in a different package):

```hs
main = do
    options <- parseOptions =<< getArgs
    grammar <- try $ runFrontend (frontendOpts options)
    (action, goto, lr1Items, unused_rules) <- runMiddleend (middleendOpts options) grammar
    case backend options of
      Normal opts -> runBackend opts grammar action goto
      RAD opts -> runRADBackend opts grammar action goto lr1Items unused_rules
```

This seems like an acceptable maintenance overhead for extension developers, compared to having to closely follow the `happy` GH repository.

int-index:
This way we can also make `happy` and `happy-boot` into separate executables, replacing the complicated flag-based logic.

[...]

PR #191, opened 2021-05-31
Modularize happy

knothed:
Following https://github.com/simonmar/happy/issues/167#issuecomment-780591344 and #187, this PR aims to split up `happy` into several components. Two main goals are accomplished:

- Introducing a clear and sensible separation between different parts of `happy` like frontend, middleend and backend.
- Making it easy for anyone to extend `happy` or to create and distribute new, possibly experimental, `happy`-features and -components (like a TemplateHaskell-frontend or a recursive ascent-descent-backend).

Ideally, `happy` programmers should be able to combine `happy` packages almost arbitrarily: Each frontend should, in theory, be able to work with each middleend, which in turn should work with each backend.

To achieve this goal, we (@sgraf812 and me) have decided to split `happy` into **six packages** – five libraries and one executable – as follows:

- `happy-frontend` is responsible for reading .y-files and parsing them into the `Grammar` datatype. 
- `happy-middleend` transforms the `Grammar` into an `ActionTable` and a `GotoTable`.
- `happy-backend` takes the these tables and generates template-based Haskell code.
- `happy-core` defines the interfaces (IRs) between the frontend, middleend and backend packages. Also, it contains core utilities like option parsing.
- `happy-test` hosts the test files and contains logic to execute the test suite.
- `happy` itself just puts the above packages together and contains little additional logic.

### Why six packages?

We want users to be able to combine different frontend-, middleend- and backend-packages together.  Therefore, `happy-core` is vital as it defines data types that every such package should be able to work with; in particular `Grammar`, `ActionTable` and `GotoTable`.

We'll look at `happy-test` shortly.

### How does a package look like?

The three packages `happy-frontend`, `happy-middleend` and `happy-backend` are all built similarly: they have a public `runFrontend`/`runMiddleend`/`runBackend` function which performs the package's main task.
Also, they have a common interface for providing and parsing command line arguments.

Let's take a look at `happy–backend` ([packages/backend](https://github.com/piknotech/happy/tree/modularization/packages/backend)). It has the following exposed components:

1. A main entry point function, `runBackend`, which performs the code-gen:

   ```haskell
   runBackend :: BackendArgs -> Grammar -> ActionTable -> GotoTable -> IO ()
   
   data BackendArgs = BackendArgs {
     outFile :: String,
     templateDir :: Maybe String,
     magicName :: Maybe String,
     strict :: Bool,
     ghc :: Bool,
     coerce :: Bool,
     target :: Target,
     debug :: Bool,
     glr :: Bool,
     glrDecode :: Bool,
     glrFilter :: Bool
   }
   ```

   `BackendArgs` are user-provided arguments, while `Grammar`, `ActionTable` and `GotoTable` come from the frontend and middleend (as results of `runFrontend` and `runMiddleend`).

   `happy` could call `runBackend` with arbitrary `BackendArgs`. There is no requirement for a command line interface per se.

2. But ultimately, we do want to perform `happy` via a CLI, meaning we want to create `BackendArgs` from parsing the command line arguments. Just as previously in `happy`, we define the CLI arguments using GetOpt, but on package-level:

   ```haskell
   data BackendFlag =
       OptOutputFile String |
       OptTemplateDir String |
       OptMagicName String |
   		...
   
   backendOptions :: [OptDescr BackendFlag] = [
       Option "o" ["outfile"] (ReqArg OptOutputFile "FILE"),
       Option "t" ["template"] (ReqArg OptTemplate "DIR"),
       Option "m" ["magic-name"] (ReqArg OptMagicName "NAME"),
       ...
   ]
   
   parseBackendFlags :: [BackendFlag] -> String -> IO BackendArgs
   ```

So, each package defines its own CLI `Flags` and `options` and, in addition, a `parseFlags` function which converts these CLI flags into `PackageArgs`.

Then, option parsing takes place as follows:

1. The main `happy` executable sticks the `options` from each package together and calls `getOpt`.
2. The resulting `Flag`s are then used to call `parseFlags` and execute each package's respective main entry point function: `runFrontend`, `runMiddleend` and `runBackend`.

In this way, option parsing is fully modularized such that new packages can define and use their own CLI options.

We just have to pay attention to one thing: with getOpt, it is not possible to have multiple options with the same option character or option string. This means, if there are two different packages which both define an `-o` option, these cannot be used together in the same `happy` executable. Package authors have to watch out for such option collisions.

### Example: happy-rad

As an example for how this modularization can be used to create a new executable, let's take a look at [happy-rad](https://github.com/knothed/happy-rad). `happy-rad` extends `happy` by a [recursive ascent-descent](https://pp.ipd.kit.edu/uploads/publikationen/knothe20bachelorarbeit.pdf) (RAD) backend. It contains two packages:

- `rad-backend`: a backend which produces RAD code. This package is completely independent of `happy-backend`. Still, it works with the same data types (`Grammar`, `ActionTable`, `GotoTable`) as `happy-backend`.
- `happy-rad`: it depends (a.o.) on `happy-frontend`, `happy-middleend`, `happy-backend` (which are remote packages) and `rad-backend` (local package). First, frontend and middleend are executed normally. Then, depending on whether the `--rad` option is set, either `rad-backend` or `happy-backend` is invoked.

As `happy-frontend`, `happy-middleend` and `happy-backend` are (will be) remote packages, we can just reuse their functionality. In addition, we get testing for free:

#### testing happy and happy-rad

How does the `happy-test` package help package authors in testing their packages?

We've moved all of the existing 26 test grammars (e.g. ParGF.y) into `happy-test`. `happy-test` exposes the following function:

```haskell
test :: TestSetup -> IO a

data TestSetup = TestSetup {
  happyExec :: String, -- e.g. happy or happy-rad
  defaultTests :: [String],
  customTests :: [String],
  customDataDir :: String,
  allArguments :: [String],
  stopOnFailure :: Bool
}
```

In their test suite, the package author can call `test` and customize the testing procedure: they can specify which of the 26 grammars should or should not be tested and which argument combinations their executable should be tested with. In addition, they can specify further, custom, testing grammars.

For example, `happy-rad` could provide `[--rad]` for the set of arguments it should be tested with, while normal `happy` could provide `[-a, -ag, -cg, -acg]`.

`happy-test` can also do something different: The functionality of `make sdist-test`, which was previously in the Makefile, has now moved into `happy-test`. This way, every extension author can test their sdists. What does `make sdist-test` do?

1. call `cabal sdist all` to create an sdist of all **local** packages
2. unzip the sdists and move them into an umbrella directory
3. create a cabal.project referring to these local packages
4. build and test the executable inside this umbrella directory.

### More

For the end user of `happy`, not much changes: the CLI stays identical, the functionality remains the same. Just the compile time of `happy` increases a bit.

#### Template Files
As the template files are backend-specific, these are now data-files of `happy-backend`, not of `happy` itself. Just like the tests, which are data-files of `happy-test` instead of `happy` itself.

#### Bootstrapping
Bootstrapping is a process which is purely frontend-specific. Therefore, the bootstrap flag only exists in `happy-frontend`, but still exists. A next step could be to outsource the bootstrapping parser into its own package, as described in https://github.com/simonmar/happy/issues/187#issuecomment-825564616.

#### CI
Travis and appveyor used `make sdist; make sdist-test-only`. We have replaced this with our new `make sdist-test`.
`make sdist-test` currently doesn't work on MinGW.


Ericson2314:
@knothed Still trying to find the time to properly sit down and review this, but I wanted to reaffirm I'm very happy to see this PR!

packages/test/src/Shell.hs
```diff
@@ -0,0 +1,61 @@
...
+runCmdIn dir args verbose = do
+  when verbose (liftIO . putStr $ "Inside " ++ dir ++ ": ")
+  liftIO $ setCurrentDirectory dir
+  runCmd args verbose
+
+runCmd' :: [String] -> Bool -> IO Bool
+runCmd' = runShell .* runCmd
+
+runCmdIn' :: String -> [String] -> Bool -> IO Bool
+runCmdIn' = runShell ..* runCmdIn
```
cgibbard, line 49: It would perhaps be clearer to write
```
runCmdIn' dir args verbose = runShell (runCmdIn dir args verbose)
```

and then you don't need to define `(..*)` which is apparently only used in this one place.

Similarly with `runCmd'`/`(.*)`.

packages/backend/src/Backend.hs
```diff
@@ -0,0 +1,96 @@
+module Backend(BackendArgs(..), Target(..), runBackend) where
```
cgibbard, line 1: Given that we're reorganising all the modules of Happy anyhow, it might be a good idea to give them all hierarchical names while we're at it. How do people feel about just putting them under the `Happy.*` hierarchy like `Happy.Backend` etc?

knothed: Sounds good, done.
Due to the separation into different cabal packages, it is unfortunately not possible to collect all the source files in one central `src/Happy` directory, which contains `Core`, `Backend`, `Frontend`, `Middleend` and `Test` subfolders. Cabal requires a separation of the packages, leading to a `packages/backend/src/Happy/Backend/...` structure; each package still has its own separate `src` directory.

packages/frontend/src/Happy/Frontend/AbsSyn.lhs
```diff
@@ -45,23 +45,17 @@ Here is the abstract syntax of the language we parse.
...
 
 ToDo: find a consistent way to analyse all the directives together and
 generate some error messages.
 
-> data ErrorHandlerType
->   = ErrorHandlerTypeDefault
->   | ErrorHandlerTypeExpList
+> data ErrorHandlerType'
+>   = ErrorHandlerTypeDefault'
+>   | ErrorHandlerTypeExpList'
```
sgraf812, line 58: Why primed?

packages/frontend/src/Happy/Frontend/Mangler.lhs
```diff
@@ -413,6 +284,10 @@ So is this.
 
 > checkRules [] _ nonterms = return (reverse nonterms)
 
+> getErrorHandlerType' :: [Directive t] -> ErrorHandlerType
+> getErrorHandlerType' = conv . getErrorHandlerType where
+>   conv ErrorHandlerTypeExpList' = ErrorHandlerTypeExpList
+>   conv ErrorHandlerTypeDefault' = ErrorHandlerTypeDefault
```
sgraf812, line 290: Ah, so there seems to be another type `ErrorHandlerType`. Why? Also `getErrorHandlerType'` returns the type without prime, while `getErrorHandlerType` returns the type with prime... Unexpected

packages/test/data/monad002.ly
```diff
@@ -135,7 +136,7 @@ The following functions should be defined for all parsers.
 This is the overall type of the parser.
 
 > type Parse = P Exp
-> calc :: Parse
+> -- calc :: Parse
```
sgraf812, line 139: Why this change?

packages/test/src/Happy/Test.hs
```diff
@@ -0,0 +1,86 @@
...
+  if result then exitSuccess else exitFailure
+
+testCons :: [(String, String, String)] -> String -> Bool -> IO Bool
+testCons [] _ _ = return True
+testCons ((dir, file, args):rest) happy stopOnFail = do
+  result <- runSingleTest happy args dir file
+  if result then testCons rest happy stopOnFail
+    else if stopOnFail
+      then return False
+      else do _ <- testCons rest happy stopOnFail; return False
```
sgraf812, line 37: The "Cons" suffix is a bit misleading. In Haskell, "cons" usually has something to with the list constructor `(:)`. The terminology is borrowed from LISP, where this constructor is even called `cons`.

packages/test/src/Happy/Test/SDist.hs
```diff
@@ -0,0 +1,118 @@
+module Happy.Test.SDist(sdist_test) where
```
sgraf812, line 1: I didn't review this properly, because I'm not familiar with how it used to work. All I know is that you spend quite a lot of time making sure that this is on par with what the Makefile used to do. I have good faith that that's the case.

packages/test/src/Happy/Test/Shell.hs
```diff
@@ -0,0 +1,55 @@
...
+
+import System.Directory
+import System.Process
+import System.Exit
+import Control.Monad
+import Control.Monad.Trans.Maybe
+import Control.Monad.IO.Class
+import Control.Applicative
+
+type TypedShell a = MaybeT IO a
```
sgraf812, line 11: This is all a bit ad-hoc and could probably take inspiration in e.g. [`shelly`](https://hackage.haskell.org/package/shelly) or [`turtle`](https://hackage.haskell.org/package/turtle). Then again, if it works it's probably good enough. Although we already had experiences that were hard to debug...

sgraf812:
I didn't review much code, because I trust you on not having modified the core happy logic. Plus, the tests seem to pass, so your refactoring seems to be successful.

I quite like the new package organisation, but ultimately it's up to the maintainers to decide.

Ericson2314:
For CI maybe try GitHub workflows. See Alex, which has made the switch, though I think is prefer a manually written one like I attempted in https://github.com/simonmar/alex/pull/190 as last I checked haskell-ci didn't support macOS with GitHub workflows yet.

Ericson2314:
Ah just merge in https://github.com/simonmar/happy/pull/196 and get it working with what you got. Hopefully we can merge that one soon and fix CI so this can be safe to merge.

.appveyor.yml
```diff
@@ -27,12 +27,9 @@ install:
  - "cabal --version"
  - "ghc --version"
  - "cabal %CABOPTS% v2-update -vverbose+nowrap"
- - "cabal %CABOPTS% v2-install alex --bindir=/hsbin"
```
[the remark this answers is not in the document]
knothed: We didn't need `alex` to begin with. Not quite sure why it was in the CI config.
But it doesn't matter when switching to GitHub Actions.

.travis.yml
```diff
@@ -17,12 +17,17 @@ env:
...
 script:
- - make sdist
- - make sdist-test-only
+ - | 
+   if [[ $(ghc --numeric-version) == "8.10.1" ]] # only perform make sdist-test once
+   then
+     make sdist-test
+   else
+     cabal test -f -bootstrap
+   fi
```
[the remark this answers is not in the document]
knothed: Yes, I only want to perform it once due to the longer running time.

packages/test/data/monad002.ly
knothed: The explicit type annotation did conflict with my rad-parser, which would add a type annotation itself when called with `--types`. I'm not sure if that's reason enough to really remove it, but since the type annotation isn't required for the test to succeed under normal `happy`, I did it.

knothed:
> Edit: Although, I wonder: Did CI really run the tests? I can see no output difference between https://travis-ci.org/github/simonmar/happy/jobs/774274988 (where they should run) and https://travis-ci.org/github/simonmar/happy/jobs/774274987 (where they shouldn't run).

Yes. In the second link, the tests have been run normally, while in the first link, `make sdist-test` was performed. Compare e.g. the number of lines from the output.

packages/frontend/src/Happy/Frontend/Mangler.lhs
knothed: You're right, the `ErrorHandlerType'` type was useless, I introduced it to break a dependency cycle, but that's actually not required.

.appveyor.yml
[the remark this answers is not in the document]
sgraf812: OK

sgraf812:
> Ah just merge in #196 and get it working with what you got

Hmm... Isn't GHA-based CI an entirely different issue? As I see it, Travis CI seems to work with this PR: https://travis-ci.org/github/simonmar/happy/builds/774274977. And it indeed runs the tests.

happy.cabal
```diff
@@ -112,100 +105,39 @@ extra-source-files:
...
                  array,
                  containers >= 0.4.2,
-                 mtl >= 2.2.1
-                     -- mtl-2.2.1 added Control.Monad.Except
+                 mtl >= 2.2.1,
+                 transformers >= 0.5,
+                 happy-core == 1.21.0,
+                 happy-frontend == 1.21.0,
+                 happy-middleend == 1.21.0,
+                 happy-backend == 1.21.0
```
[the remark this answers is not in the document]
sgraf812: We could use `configure` to do the job and also check in the generated `cabal` files. Alternatively just put up with having to grep for and replace 10ish references to the pretty unique version string `1.21.0`...

Ericson2314:
@knothed Oh when I clicked "all checks have passed" Travis was not there, so I assumed it was broken. Nevermind then.

ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## 1.21.0
```
Ericson2314, line 1:
```suggestion
## Unreleased - 1.21.0
```
Reminds others to add things here and not in new section :)

.appveyor.yml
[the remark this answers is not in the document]
Ericson2314: I can resolve it, button isn't greyed out.

.travis.yml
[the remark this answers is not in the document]
Ericson2314: What makes it longer running, out of curiosity?

happy.cabal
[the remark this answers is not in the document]
Ericson2314: > generated `cabal` files

Hmm? What is that?

packages/test/data/monad002.ly
Ericson2314:
```suggestion
> calc_test_sig :: Parse
> calc_test_sig = calc

(We use a fresh name because `calc` might already have a sig depending on the backend.)
```

packages/test/src/Happy/Test/Shell.hs
Ericson2314: Yeah as long as this doesn't need to change too often I don't care too much.

Ericson2314:
OK I would like this to get merged soon :). It is very good and has sat open too long.

A few more things

1. Can we put `happy` itself in `packages/` too? This will make filtering the sources per package from the repo easier, because every package gets its own subtree.
2. I would be open to giving the swappable packages names based on the algorithm they currently use. e.g. maybe we would switch the default backend from this one to a different one at some point :).
3. Can the attribute grammar stuff leverage this some how? I see you have `isBootstrapped` to leak out whether to do the attribute grammar tests.

Ericson2314:
If you check the "maintainers can push to PR" box I can help fix conflicts.

Ericson2314:
Also https://github.com/piknotech/happy/pull/1

knothed:
> OK I would like this to get merged soon :). It is very good and has sat open too long.

Nice, I'd also like that :)

> 1. Can we put `happy` itself in `packages/` too? This will make filtering the sources per package from the repo easier, because every package gets its own subtree.

Yes we could do that. This would however require all of `happy`s data-files to be inside this subdirectory. These include the readme, changelog, makefile etc., which would be much better placed in the top directory instead of being inside `packages/happy`.
We could probably solve this using links? But I don't know whether that is cross-system-proof.

> 2. I would be open to giving the swappable packages names based on the algorithm they currently use. e.g. maybe we would switch the default backend from this one to a different one at some point :).

Sounds good - what would you propose?

> 3. Can the attribute grammar stuff leverage this some how? I see you have `isBootstrapped` to leak out whether to do the attribute grammar tests.

Bootstrapping is a frontend-specific thing, and the frontend exposes `isBootstrapped`. The other packages don't care about bootstrapping.

Using `happy` with attribute grammars requires the bootstrapped `happy` however which is why the attribute grammar tests are listed separately inside `Tests.hs`.

Ericson2314:
> We could probably solve this using links? But I don't know whether that is cross-system-proof.

I'm told windows support for symlinks is continuously improving. And I think git had some tricks even before that, so that's fine with me.

Ericson2314:
> > 1. I would be open to giving the swappable packages names based on the algorithm they currently use. e.g. maybe we would switch the default backend from this one to a different one at some point :).
> 
> Sounds good - what would you propose?

I guess first I would need to understand how/why your happy rad still uses regular `happy-backend`.

Ericson2314:
Well `isBootstrapped` should be called something like `supportsAttributeGrammars` for starters. The former is an implementation detail, the latter isn't.

knothed:
> I guess first I would need to understand how/why your happy rad still uses regular `happy-backend`.

My rad-backend is independent of `happy-backend`. The executable `happy-rad` however does include regular `happy-backend` so the user can decide which backend they wanna use.

knothed:
Should we remove the TODO file? It seems outdated.

Ericson2314:
> > I guess first I would need to understand how/why your happy rad still uses regular `happy-backend`.
> 
> My `rad-backend` is independent of `happy-backend`. The executable `happy-rad` however does include regular `happy-backend` so the user can decide which backend they wanna use.

Oh! So it could be skipped to force them into the rad backend?

Ericson2314:
Could you make a `happy-backend-glr`? And make it optional just like you made `happy-rad` support two backends?

Perhaps that is a better splitting than whatever I was thinking of with the attribute grammars :).

knothed:
Yes, I‘ve also thought about splitting the GLR backend off. I‘ll do that.

It could however be that the GLR backend  has a dependency to the normal backend, as it uses, I think, some of the normal backend logic. But that shouldn’t be a problem – the normal backend (obviously) doesn’t depend on the GLR backend.

knothed:
> Oh! So it could be skipped to force them into the rad backend? That's great news.

Exactly!

Ericson2314:
https://github.com/simonmar/happy/blob/1ad42c11b24ca6bc67720489a21fc7e0aa23dcfe/packages/backend/src/Happy/Backend/ProduceGLRCode.lhs actually looks like it just uses core!

Ericson2314:
It seems the attribute grammar *does* have a backend component, after all, or at least there is some ugly string concatenation in the frontend mangler that look like it.

But this is no fault of the current PR, so I'm happy to deal with it post merge :)

knothed:
`cabal check` wants license and maintainer fields for each package. I've just copied most fields from the original happy package, if that's okay.

I've also added some imperfect descriptions to the packages, maybe you could take a quick look at them.

knothed:
GitHub actions works now (https://github.com/piknotech/happy/actions). I've manually updated `haskell-ci.yml` and `haskell-ci.patch` as `haskell-ci` doesn't seem work on Mac with GitHub workflows, so someone should probably test the `haskell-ci.yml` file generation again.

Also, could you approve the GitHub actions to run here in this repo? As a first-time contributor I need a maintainer to approve running workflows.

knothed:
> `cabal check` wants license and maintainer fields for each package. I've just copied most fields from the original happy package, if that's okay.

Maybe this would be an opportunity to name the new maintainers, as currently only Simon Marlow is present in the cabal files.

.travis.yml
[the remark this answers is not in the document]
knothed: `make sdist-test` builds and tests happy twice (non-bootstrapped and bootstrapped), `cabal test` only once.

knothed:
@Ericson2314 The GLR-backend is now split off into a separate package.
It is however (and has previously not been) covered by any tests.

Are you happy with the cabal files?

knothed:
`happy-rad` now even features three backends at once: normal, glr and rad. It calls the correct one depending on whether the `--glr` or `--rad` flags are set.

This multi-package option parsing can be seen [here](https://github.com/knothed/happy-rad/blob/master/happy-rad/src/Main.hs). Could be helpful for future package developers.

packages/backend-glr/src/Happy/Backend/GLR/CLI.hs
```diff
@@ -0,0 +1,58 @@
+module Happy.Backend.GLR.CLI(Flag(..), options, characteristicOption, hasCharacteristicFlag, parseFlags) where
+
+import Prelude hiding (filter)
+import Happy.Backend.GLR
+import System.Console.GetOpt
+
+-------- CLI flags and options --------
+
+data Flag =
+    OptGLR |
```
Ericson2314, line 10: @knothed actually I think this should be a top level flag. The top level decides what backend to to use. Once we are in this module we've already made that this, and this package doesn't use this flag at all.

.travis.yml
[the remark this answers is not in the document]
Ericson2314: OK, I should make separate jobs for with and without bootstrap to do in parallel after.

Ericson2314 (approved):
@knothed baring the last thing I mentioned above, this has my approval! Great work!

I do want to hear from @int-index or @simonmar so I'm not doing something so big unilaterally, but that's it!

tests/error001.stderr
```diff
@@ -1,5 +0,0 @@
-error001.y: Multiple rules for 'foo'
```
Ericson2314, line 1: Oh sorry, I forgot to ask: shouldn't this go somewhere?

int-index:
I love the idea, but this is impossible to review. Can you split it into several parts?

You could first factor out `happy-core` into its own package as a separate PR, then `happy-middleend`, then `happy-backend`, etc, each in its own incremental PR.


Also, I disagree a little with the way you have split this into packages, in particular regarding to `happy-core` and `happy-middleend`. Command-line options are a part of the command-line interface and definitely do not belong to `happy-core`, whereas routines to transform `Grammar` into `(ActionTable, GotoTable, [Lr1State], [Int])` can be easily put into the same package as `Grammar`, `ActionTable`, and `GotoTable`.

So I think that:

* utilities related to command-line parsing go to `happy-cli` instead of `happy-core`
* functionality of `happy-middleend` should be merged into `happy-core`, and the term “middleend” should be avoided altogether (I’d rename `runMiddleend` to `buildTables` or something)

Ericson2314:
@int-index I do not like the principle of putting CLI stuff everywhere either, but it does seem awfully convenient for whipping up exes like @knothed's `happy-rad`

int-index:
>  it does seem awfully convenient for whipping up exes like @knothed's happy-rad

True, hence the suggestion to put CLI utilities into their own package `happy-cli`

Ericson2314:
> > it does seem awfully convenient for whipping up exes like @knothed's happy-rad
> 
> True, hence the suggestion to put CLI utilities into their own package `happy-cli`

So it's OK if happy-frontend and happy-backend depend on those?

int-index:
> So it's OK if happy-frontend and happy-backend depend on those?

No. What is there in `packages/core/src/Happy/Core/OptionParsing.hs` (in the current patch) that the frontend or the backend would use?

Ericson2314:
Oh, I see `packages/*/src/Happy/*/CLI.hs` doesn't actually use that stuff.

Ericson2314:
Anyways, I like the `happy-cli` idea, it doesn't resolve the tensions but it is an improvement.

I would like to keep `middleend` separate, however. It is a bad name, but I do think it is useful to separating a yet-to-be-named middle stage of compilation from the stuff everything needs.

Ultimately, I don't think happy is written with modularity in mind, so it's better to do *some* split, erring on the side of more granularity, and then refactor things to be sane after. (I do not want to do a release right after this PR is merged.)

int-index:
> Are you OK with that stuff being there since it is still morally CLI stuff even if it doesn't use that module?

Are you referring to `packages/*/src/Happy/*/CLI.hs`? No, I don’t think it belongs there either. But it’s more tolerable, because there’s no immediate use case that would suggest doing otherwise.

I plan to use `happy-core` as part of my TTH-based EDSL, and it definitely has nothing to do with CLI. On the other hand, something like `happy-backend` would only be used without CLI if we planned to support a Happy GUI or something, which is less likely. (But if it happens – we will need to think about moving CLI stuff out).

> I do think it is useful to separating a yet-to-be-named middle stage of compilation from the stuff everything needs.

Okay, but let’s think about concrete use cases. What library or app would ever require `Grammar` and `ActionTable` / `GotoTable` but not the function to transform from one to another?

Ericson2314:
We're on same page for the CLI stuff now.

For the Grammar, I thought one could convert a Grammar to the tables and then dispense with the grammar, but the imports seem to indicate otherwise. That pulls the foundation out from underneath that the middle stuff is a pipeline stage at all.

(To recap, the structure I generally like would be ASTs/IRs 0..n+1, and transformations 0..n, depending online on the input and output ASTs/IRs.)

sgraf812:
FWIW, initially we thought about a split of just frontend > middleend > backend and middleend was all about turning Grammar descriptions into ActionTables. But then we discovered that all of them need CLI functionality and share types. It would've been awkward to put them all in frontend, so we wrote core to put the shared stuff there.

Now it might seem that middleend doesn't do much and we could move it into core, but then frontend and middleend implicitly depend on how we translate Grammars to ActionTables... Maybe that is OK?

CLI stuff ended up in core because we didn't want to introduce yet another package, but happy-cli is ultimately the better solution.

Ericson2314:
OK it seems all the backends really do use the original `Grammar` meaningfully.

If that is something that is reasonable to fix (by instead passing in the residual information that is not duplicate with that in tables) I would have `happy-grammar` and `happy-tabular` together with `happy-cli` replace `happy-core`.

The final breakdown would be:

`Tables` -> `happy-tabular`
`GenUtils` -> `happy-grammar`, but should be liquidated
`Grammar` -> `happy-grammar`
`NameSet` -> inline to `happy-middleend`, redefine in happy-rad
`OptionParsing` -> `happy-cli`

Backends would depend on `happy-grammar` and `happy-tabular` until that situation is rectified.

If, on the other hand, it is not reasonable to keep `Grammar` away from the backends, then I would further combine `happy-middleend` into `happy-tabular`. (The backends would morally depend on how the tables were projected, but the fronted wouldn't depend on the tables.)

packages/backend-glr/src/Happy/Backend/GLR/CLI.hs
knothed: True, done.

I've also thought about this, but wasn't sure as it requires every executable that uses multiple backends to declare these backend-switching flags itself, leading to code duplication across executables.

But that shouldn't really be a problem.

tests/error001.stderr
knothed: Yes, but I hesitated to reproduce the logic for running error-tests (i.e. test that are expected to fail) and comparing their output with some previously defined output, given that there is only a single one of these error-tests which is additionally very trivial.

I believe it it okay to drop `error001.y`.

Should more of these error-tests be written, `happy-test` can be extended accordingly.

Ericson2314:
> All of happy-middleends CLI-options would just be merged into the frontend, and the frontend would perform the IO-tasks that the middleend did so far: info file generation and debug dumping.

I wouldn't do that, because then happy-frontend gains the happy-tabular dep. I would just leave those in happy-tabular for now, and we can move them out later.

> What would you suggest to improve the situation with the CLI? 

Fine as is. The individual packages don't use that CLI module that was in core, so we side-step it for now. When we do the TH edsl we can consider doing the -cli package explosion as @int-index implied (as I interpreted, at least).

cdsmith:
Not a productive comment, but I just wanted to say that I'm really happy to see this, as I was looking for a happy parser that I could use to partially automate a labor-intensive conversion from GHC's Haskell grammar.  This looks like just what I need.  I'm quite eager to see it merged.

Ericson2314:
@cdsmith On the contrary it's great motivation to here this stuff doesn't just feel like a good idea, but has concrete immanent use-cases something laudable like code world (I presume :)).

Ericson2314:
Welcome back, @knothed :). @int-index what do you think of my last proposal? I was actually expecting us to decide on something first before @knothed implemented it — hopefully it's more too many more hoops to jump through!

int-index:
Yeah, it’s fine by me.

But as I said, this current PR changes too much at once, I find it difficult to review. Could we do it gradually? For example, start by separating out `happy-grammar` into its own package.

knothed:
Perfect, I’ll try splitting up the MR as sensibly as possible. :)

Would you like each MR to actually build and work for itself? That would be a bit challenging considering the changes to the parsing architecture.

int-index:
> Would you like each MR to actually build and work for itself?

Yes.

> That would be a bit challenging considering the changes to the parsing architecture.

All the more reason to do things separately. Maybe we should start with changes to the parsing architecture without moving things around -- that would make it easier to review the new architecture as a proper diff

Ericson2314:
I suppose the first PR can establish the subdirectory regime (the moves will not clutter the diff), and then the subsequent ones are more simple.

knothed:
Hmm, wouldn’t moving the source files into different directories break the single-package `happy`, which expects them in `src/`?

Ericson2314:
I just mean the first PR can adopt the directory structure of this PR, but with just the two packages happy and happy-grammar.

knothed:
I decided for the following path:
1. Split off `happy-grammar` and `happy-tabular`
2. Split off `happy-frontend`, `happy-backend` and `happy-backend-glr`
3. Create `happy` and `happy-cli`
4. Create `happy-test`.

Changing all the CLI stuff at once and at the end is the most convenient, both for doing so and for reviewing.

@int-index The first MR is #200 which can be reviewed now.

Maybe it makes more sense to merge all the MRs into an intermediate branch and then merge that branch into master only at the end. Maybe someone could create a branch (e.g. `modularisation`).

PR #196, opened 2021-07-14
CI on GitHub via `haskell-ci github happy.cabal`

### LINKED ISSUE #189: Switch to Github Workflows for CI?
I hate to pull all our eggs in on one basket called GitHub, but Travis has been rather slow, and it would be nice to share more stuff between Unix and Windows CI.

Maybe the first step is to have all 3 and see how they perform in practice?

I do not have enough repo permissions, however, to try this myself.

### LINKED ISSUE #193: travis-ci.org shutdown
happy currently uses www.travis-ci.org, which will be shut down on June 15, i.e. very soon.

We should migrate to **www.travis-ci.com** to further use travis.

Once #189 is implemented, travis could also be dropped completely.

andreasabel:
Out-of-the box CI on GitHub using the haskell-ci tool, put patched for happy, similar to https://github.com/simonmar/alex/pull/189.

- `DEVELOPER.md` explains how to regenerate the CI, e.g. in case of new GHCs.

- `.github/haskell-ci.patch` modifies the generated workflow by:
  * Setting flag 'bootstrap' to False in calls to cabal.
  * Installing happy instead of just building it.
  * Setting HAPPY and HC for running the tests.

- `tests/Makefile` has been modified to not overwrite already set `HAPPY` and `HC` variables.  This is needed to locate these executables correctly.

- `cabal.haskell-ci` contains configuration for running `haskell-ci`:
  It removes constraints on packages `containers`, `mtl` and `transformers`.
  Otherwise, these will be constrained to the GHC-shipped versions, which isn't compatible with `happy` for all GHC versions.

[fixes #189] [fixes #193]


andreasabel:
GHA workflow run is at https://github.com/andreasabel/happy/actions/runs/1031280160

Ericson2314:
Can you delete Travis too?

Ericson2314:
I emailed @simonmar and @int-index about enabling GitHub actions in this repo.

andreasabel:
> Can you delete Travis too?

Avec plaisir.  Voila!

> I emailed @simonmar and @int-index about enabling GitHub actions in this repo.

Good.  I do not fully understand how this works, but it could be that the actions run as soon as they are in master...

Ericson2314:
Oh right, that might be how it was before. Well, taking the risk :).

andreasabel:
> Oh right, that might be how it was before. Well, taking the risk :).

Great, that worked: https://github.com/simonmar/happy/actions/runs/1050964788

PR #201, opened 2021-10-01
Replace 'optIO' with 'when'

int-index:


src/Main.lhs
```diff
@@ -297,13 +297,10 @@ Successfully Finished.
 > dieHappy :: String -> IO a
 > dieHappy s = getProgramName >>= \prog -> die (prog ++ ": " ++ s)
 
-> optIO :: Bool -> IO a -> IO a
-> optIO fg io = if fg then io  else return (error "optIO")
```
Ericson2314, line 301: Oh my... I thought that `error "optIO"` meant this was some sort of assert... and then I realized it was just some nonsense to make the type checker shut up when the type checker is actually right.

Someone needs to write up how the earlier history of Haskell was really not at all "well typed programs don't go wrong" philosophy.

PR #202, opened 2021-10-01
Debug options in all build configurations

int-index:
Hiding this behind CPP is not necessary. GHC supports flags such as `-ddump-tc-trace` and users have no problem with this.


Ericson2314:
Good call. Especially as we are making libraries, should always have the `Show` instance.

PR #203, opened 2021-10-01
Create the 'Tables' data type

int-index:
This data type encapsulates the result of processing the `Grammar`.

In addition to `ActionTable` and `GotoTable`, it also includes intermediate results such as `Lr0Items`, which are needed for debug printing and info table generation.

The `find_redundancies` function has been moved from `Main.lhs` to `Tabular.lhs` almost verbatim (use `git show --color-moved=blocks` to confirm).


Ericson2314:
I definitely like `genTables` over `runTabular`, but do we want to go from the function param to the `glr` boolean? I thought `happy-tabular` was supposed to be somewhat agnostic to the parsing strategy --- must maybe that's a impossible/nonsensical ideal so the `glr` bool is fine?

int-index:
Sure, I can revert that bit.

Ericson2314:
Thanks. I took a look at https://github.com/simonmar/happy/pull/200 to confirm and indeed the GLR stuff was pretty well separated in the end, so that goals does look attainable.

int-index:
Well, #200 uses a `Bool` flag too, I used it as a reference.

Ericson2314:
@int-index Ah I see. Yes, you were very right to demand we make the diffs readable. These things slipped past me.

Ericson2314:
Also, btw, I am rebasing #200 on master and then this to queue up the next steps, so you can just worry about this for the moment.

src/Tabular.lhs
```diff
@@ -0,0 +1,93 @@
...
+>           conflicts   = {-# SCC "Conflict" #-} (countConflicts actionTable)
+>           redundancies = find_redundancies select_reductions g actionTable
+>       in Tables { lr0items, la_spont, la_prop, lookaheads, lr1items,
+>                   gotoTable, actionTable, conflicts, redundancies }
+
+-----------------------------------------------------------------------------
+Find unused rules and tokens
+
+> find_redundancies
+>        :: (LRAction -> [Int]) -> Grammar -> ActionTable -> ([Int], [String])
```
Ericson2314, line 53:
```suggestion
>        :: SelectReductions -> Grammar -> ActionTable -> ([Int], [String])
```

int-index: Yeah, I couldn’t decide whether to use the type synonym or to keep the type signature verbatim when copying.

Alright, let’s change this one :-)

Ericson2314 (approved):

knothed:
Nice! So do we want to merge #200 on top of this?

int-index:
Definitely some parts of #200, since it introduces the `.cabal` files and so on.

knothed:
Okay, will you and @Ericson2314 do this?
I would then after #200 is merged follow up with the next MR concerning the frontend and backend.

Also, I‘d like to make some more grammar-related types derive `Show`, and I’d appreciate it if `Grammar` would contain `hd` and `tl` fields so we wouldn’t always have to pass them separately. Can we integrate this in #200 or should this be a follow-up MR?

Ericson2314:
@knothed see what I wrote in https://github.com/simonmar/happy/pull/200#issuecomment-932634187. I am adding a little checklist to that PR.

PR #207, opened 2021-10-16
Update readme, mainly with build instructions

Ericson2314:
Now I must admit I'm getting cold feet and feel like I should maybe see how hard the cabal-install issue to fix is, but it is a time-intensive investigation

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,15 +9,35 @@ Happy is a parser generator for Haskell 98 (and later).
 
 ## Build Instructions
 
-Happy is built using Cabal.  First install GHC, then:
-```
-  $ cabal install
-```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,15 +9,31 @@ Happy is a parser generator for Haskell 98 (and later).
...
+If you *do* have an existing Happy executable on the PATH, do regular
 ```
-  $ cabal install
+$ cabal build
 ```
-If you obtained the development version from https://github.com/simonmar/happy/,
-install via:
+like with any other project.
+
+If you do *not* have an existing Happy executable, instead do
```
int-index, line 22: This can be one command `cabal build -f -bootstrap`

PR #209, opened 2021-10-20
Create Tabular package

knothed:
Create a Tabular package which contains the `Tabular` module and related stuff like LALR.


packages/grammar/src/Happy/Grammar.lhs
```diff
@@ -123,4 +125,24 @@ In hindsight, this was probably a bad idea.
...
 > errorTok        = 1
-> epsilonTok      = 0
\ No newline at end of file
+> epsilonTok      = 0
+
+-----------------------------------------------------------------------------
+Replace $$ with an arbitrary string, being careful to avoid ".." and '.'.
+
+> mapDollarDollar :: String -> Maybe (String -> String)
+> mapDollarDollar code0 = go code0 ""
```
Ericson2314, line 134: How come this is moved here? It seems only used by backend code? I think we should leave in genutils at least temporarily, unless I am missing something.

packages/tabular/src/Happy/Tabular/Info.lhs
```diff
@@ -4,14 +4,13 @@ Generating info files.
...
+> module Happy.Tabular.Info (genInfoFile) where
 
-> import Paths_happy            ( version )
-> import LALR                   ( Lr0Item(..), Goto(..), LRAction(..), ActionTable, GotoTable )
-> import GenUtils               ( str, interleave, interleave' )
 > import Data.Set ( Set )
 > import qualified Data.Set as Set hiding ( Set )
 > import Happy.Grammar
+> import Happy.Tabular.LALR   ( Lr0Item(..), LRAction(..), Goto(..), GotoTable, ActionTable )
+> import Paths_happy_tabular  ( version )
```
Ericson2314, line 13: I wonder if we should skip the

>   showHeader
>       = banner ("Info file generated by Happy Version " ++
>                 showVersion version ++ " from " ++ filename)

and keep that in the binary for now. I feel a bit weird embedding a library version number like that.

Ericson2314:
Oh sorry I lost track of this. I am fine with `Lrt1State` for now, and I think I am fine with the rest of this too. @int-index you have any thoughts?

Ericson2314:
#200 I rebased to just be additional splits and the main part, so I am happy to keep it around for later for when we revisit that sort of stuff :)

int-index:
I will review by Monday.

int-index:
I have not reviewed by Monday. Shame on me. But I’m doing it right now.

int-index:
LGTM. Thank you!

PR #212, opened 2021-10-21
Create `happy-boot` exe

Ericson2314:
This helps avoid missing features in cabal-install, but doesn't yet work
because Cabal doesn't realize it should use `happy-boot` for `.ly`
files.


int-index:
> Cabal doesn't realize it should use happy-boot for `.ly` files

Could this be resolved by having a custom Setup.hs and calling happy-boot by hand?

phadej:
Having `build-type: Custom` is not great solution.

If I understand right, this work is to make developing experience of `happy` itself better. That **should not** come with trade-offs for `happy` users (and lets not forget that there are other users in addition to GHC).

IMHO, if parser can be generated locally it should... Until there is needed support in `Cabal` or `cabal-install`.

int-index:
@phadej OK. How do you imagine it should be done? 

Should there be a field in the `.cabal` file specifying that `.y` and `.ly` files are to be processed with `happy-boot`?

phadej:
> Would this field allow mapping arbitrary extensions to arbitrary build tools?

I'd imagine it work so. That would be great addition in general. Hardcoding/blessing just some tools is not future-proof nor extensible.

Ericson2314:
There's a small irony here that In the longer term I hope happy will always be usable via TH (even in GHC), and there will be no need for these custom preprocesors or generalized support for them in Cabal, but we're not there yet.

PR #216, opened 2021-11-24
Create Frontend package

knothed:
This PR is pretty straightforward, it moves all frontend-specific files into the `happy-frontend` package.


knothed:
@Ericson2314 @int-index

knothed:
Todo: adapt Haskell-CI (note to myself)

packages/frontend/src/Happy/Frontend.hs
```diff
@@ -0,0 +1,57 @@
...
+  where first f (Left a) = Left (f a)
+        first _ (Right b) = Right b
+
+writePrettyFile :: Maybe String -> AbsSyn -> IO ()
+writePrettyFile location abssyn = do
+  let out = render (ppAbsSyn abssyn) in
+    case location of
+      Just s   -> writeFile s out >>
+        hPutStrLn stderr ("Production rules written to: " ++ s)
+      Nothing  -> return ()
```
Ericson2314, line 26: We don't want random IO stuff in the library. This should stay in the exe.

packages/frontend/src/Happy/Frontend.hs
```diff
@@ -0,0 +1,57 @@
...
+
+possDelitify :: String -> String -> IO (String, String)
+possDelitify ('y':'l':'.':nm) fl = return (deLitify fl, reverse nm)
+possDelitify ('y':'.':nm) fl     = return (fl, reverse nm)
+possDelitify f            _      = die ("`" ++ reverse f ++ "' does not end in `.y' or `.ly'\n")
+#if !MIN_VERSION_base(4,8,0)
+  where die s = hPutStr stderr s >> exitWith (ExitFailure 1) 
+#endif
```
Ericson2314, line 34: This function isn't ready to come out of the exe yet. It cuts across concerns to weirdly.

If you want, you can make a function here to parse the filename (including doing the reversing), and it should return `Maybe Bool` or something instead of any `IO`. Then the exe can use that, do the `die`-ing, and de lit the file according to the `Bool`.

packages/frontend/src/Happy/Frontend.hs
```diff
@@ -0,0 +1,57 @@
...
+import System.Exit
+import System.IO
+
+parseYFileContents :: String -> ParseResult AbsSyn
+parseYFileContents contents = runFromStartP ourParser contents 1
+
+mangleAbsSyn :: AbsSyn -> String -> ParseResult Grammar
+mangleAbsSyn abssyn filename = first unlines (mangler filename abssyn)
+  where first f (Left a) = Left (f a)
+        first _ (Right b) = Right b
```
Ericson2314, line 18: This isn't really worth factoring out, it is just a little easy-to-write helper (especially with `Data.Bifunctor`). Better to keep that stuff as-is in the exe.

packages/frontend/src/Happy/Frontend.hs
knothed: True. The reason I put this (and `mangleAbsSyn` below) in `Frontend` is to reduce the number of exposed modules of `happy-frontend`, but yeah this shouldn't really be that relevant.

knothed:
> Todo: adapt Haskell-CI (note to myself)

Done. Seems to work.

Ericson2314 (approved):
Thanks! This looks good to me now. @int-index?

int-index:
I haven’t looked at this as carefully as at the previous patches, but overall this looks good to me.

Ericson2314:
This diff is indeed simpler to read, so I don't think there is anything we missed. I'll merge it. Thanks again, @knothed!

PR #225, opened 2022-02-24
Instructions how to build the happy documentation

andreasabel:
Instructions how to build the happy documentation.
Also, adding path to DocBook stylesheets (xsl) when installed on Mac via HomeBrew.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ cabal-dev
 cabal.sandbox.config
 .*.swp
 .*.swo
+/.vscode/
--- a/doc/configure.ac
+++ b/doc/configure.ac
@@ -1,11 +1,11 @@
 
-AC_INIT([Haddock docs], [1.0], [simonmar@microsoft.com], [])
+AC_INIT([Happy docs], [1.0], [smarlow@fb.com], [])
 
 AC_CONFIG_SRCDIR([Makefile])
 
 dnl ** check for DocBook toolchain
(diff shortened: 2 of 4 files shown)


andreasabel:
@Ericson2314 @simonmar : Can I get commit rights on the `happy` repo as well?

doc/configure.ac
```diff
@@ -1,11 +1,11 @@
 
-AC_INIT([Haddock docs], [1.0], [simonmar@microsoft.com], [])
```
Ericson2314, line 2: hah!

.gitignore
```diff
@@ -6,3 +6,9 @@ cabal-dev
 cabal.sandbox.config
 .*.swp
 .*.swo
+/doc/autom4te.cache/
+/doc/config.log
+/doc/config.mk
+/doc/config.status
+/doc/configure
+/doc/happy.1
```
Ericson2314, line 14: I would put these in a separate `.gitignore` inside the docs subdir. How does that sound?

doc/configure.ac
andreasabel: Fixed simon's email address.

andreasabel:
Btw, these instructions are ported from the alex repo.

It seems contemporary to switch the documentation to sphinx.  I started out with the Alex one, but if this is approved, I can also take a stab at the Happy one.

Ericson2314:
I support reST conversion. Good to be consistent with GHC and Cabal.

Issue #62, opened 2016-05-12
fix deadlinks to glr thesii in the docs 


cartazio:
attached is 
Pure Functional Parsing
an advanced tutorial
Peter Ljunglof
[Pure Functional Parsing an advanced tutorial Peter Ljunglof Ljunglof-2002a.pdf](https://github.com/simonmar/happy/files/262127/Pure.Functional.Parsing.an.advanced.tutorial.Peter.Ljunglof.Ljunglof-2002a.pdf)

not sure what papers / docs correspond to the other GLR units of work

PR #228, opened 2022-03-06
licentiate thesis

askeblad:
fix for https://github.com/haskell/happy/issues/62

Göteborgs Universitetsbibliotek
https://gup.ub.gu.se/publication/10783

or

Gothenburg University Library
https://gup.ub.gu.se/publication/10783?lang=en


Ericson2314:
Thanks for fixing these in the original!

Issue #197, opened 2021-07-23
Examples don't work


knothed:
Many of the examples do not work. They seem to use outdated techniques of happy and often import modules that don't exist.

askeblad:
ErrorTest.hs
LexerTest.ly
MonadTest.ly

```diff
- import Char
+ import Data.Char
```

askeblad:
Calc.ly

```diff
- import Char
+ import Data.Char
```
after running `happy Calc.ly` an error results


```
$ runghc Calc.hs

Calc.hs:0:53: error:
    * Variable not in scope: main :: IO a0
    * Perhaps you meant `min' (imported from Prelude)
$ 
```

askeblad:
```
$ happy PgnParser.ly
PgnParser.ly:26: unrecognised directive: %newline
$
```

askeblad:
DavesExample.ly

```
> import Lexer
> import Convert
> import PreludeGlaArray
```
ErlParser.hs

```
import GenUtils
import Lexer
import AbsSyn
import Types
import ParseMonad.Class
```
SimonsExample.hs

```
import Type
import Lexer
```

PR #230, opened 2022-03-07
examples update

askeblad:
if someone else wants to correct the other top level [examples](https://github.com/haskell/happy/issues/197) I can add those changes in additional commits


askeblad:
@Ericson2314

you can go ahead and merge

it might be some time before an assessment is made on the other examples

PR #247, opened 2023-01-31
Get rid of an incomplete pattern match

josephcsible:


packages/backend-glr/src/Happy/Backend/GLR/ProduceCode.lhs
```diff
@@ -202,7 +202,7 @@ Extract the string that comes before the module declaration...
 >     . table_text
 
 >   lib_content lib_text
->    = let (pre,_drop_me : post) = break (== "fakeimport DATA") $ lines lib_text
+>    = let (pre,post) = break (== "fakeimport DATA") $ lines lib_text
```
andreasabel, line 205: Should we make such a change?  I think it is hurting code comprehensibility more than helping it.
The current formulation makes clear that the `_drop_me :  post` match MUST succeed, even if GHC does not see it.
The new formulation clouds this insight.
So, I am against such a change.

Ericson2314: Yes it would be better to add an explicit pattern match with an `error ...`. It would be even better to try to use `NonEmpty`.

andreasabel: `NonEmpty` is available from `base-compat` for old GHC versions back to 7.0 (`base-4.3`), so that would not be a problem for keeping all currently supported GHC versions.

sgraf812:
Fixed the incomplete match warnings in #281.

josephcsible:
I'd argue that #281 didn't *fix* anything, but merely *hid* the warnings.

andreasabel:
Well, the problem with these warnings is that you cannot switch them on a _case-by-case_ basis.
The workflow would be:
1. The partiality warning is raised.
2. The case is investigated.
3. a. If the partiality can definitively not occur at runtime, the warning is switched off for _this case_.
   b. If the situation is unclear, the case should be fixed to catch a possible partiality.

For this issue, we are in case 3a, but unfortunately GHC only allows us to switch off the warning globally.

PR #248, opened 2023-02-06
Merge in changes from release branch

Ericson2314:
1. Ensure the the change log is up to date.

2. Fix some small markdown errors I found in the older parts of the change log.

3. Forward-port fix for newer `mtl`.

4. Get changes to Cabal metadata reflecting joint maintainership.

I like merging in release branches to keep things in sync, rather than a cherry-picking dance or just doing things purely manually.

PR #249, opened 2023-02-06
Update to the latest Haskell CI (0.15.20230203)

andreasabel:
Update to the latest Haskell CI (0.15.20230203)
- update `tested-with` in the `.cabal` files
- update the patch to apply cleanly to the latest Haskell-CI generated workflow
- update developer manual with step to update the patch
- update `haskell-ci.yml`

I split this into several patches as some of these could be applied to the `release-1.20` branch, but not all.  See:
- #248


andreasabel:
CI seems to fail because some Ubuntu server isn't available.  Need to try later again.

andreasabel:
> CI seems to fail because some Ubuntu server isn't available. Need to try later again.

CI passes now: I reran the pull_request instance of CI (succeeds), the push instance is identical (so I did not rerun it; still shows failure).

PR #256, opened 2023-10-01
Fix sdist

### LINKED ISSUE #255: Happy refactoring (happy-1.21): either move forward or take it back
ATTN: @Ericson2314 @int-index 

`happy-1.21` included large refactorings, was released in October 2021 but then unreleased shortly after.  It is soon that this is two years ago.

Unfortunately, in the current state, with refactorings applied but a boot-strapping problem persisting, `master` is not releasable.  This also means that fixing issues is stalled, as the fate of those fixes would be unknown.

I only see two options now:
1. Either finish to work done in 2021 to bring it to a releasable state.
2. Or take this work back and bring `master` back into a releasable state based on the `happy-1.20` code.

Since I am mostly interested in the stability of `happy`, I don't care much which path is taken.  However, I think that something has to happen soon.

int-index:
Fixes #255

The new release process is:

1. `runhaskell ./bootstrap.hs`. This replaces `.y` and `.ly` files with `.hs` files, which applies to:
    * `packages/frontend/src/Happy/Frontend/Parser/Bootstrapped.ly`
    * `packages/frontend/src/Happy/Frontend/AttrGrammar/Parser.ly`
2. `cabal v2-sdist all`

This results in the following tarballs:

```
dist-newstyle/sdist/happy-backend-lalr-2.0.tar.gz
dist-newstyle/sdist/happy-grammar-2.0.tar.gz
dist-newstyle/sdist/happy-tabular-2.0.tar.gz
dist-newstyle/sdist/happy-backend-glr-2.0.tar.gz
dist-newstyle/sdist/happy-codegen-common-2.0.tar.gz
dist-newstyle/sdist/happy-frontend-2.0.tar.gz
dist-newstyle/sdist/happy-2.0.tar.gz
```

Thanks to the `./bootstrap.hs` step, the `happy-frontend-2.0.tar.gz` tarball contains `*.hs` files instead of `Bootstrapped.ly` and `Parser.ly`. This should fix the issues that the users of `1.21.0` encountered.

The bump to `2.0` is motivated by the split into packages. I believe it's a major change that enables the use of `happy` as a library, so it deserves a major version bump. It does **not** indicate a breaking change.

### DIFF
--- a/bootstrap.hs
+++ b/bootstrap.hs
@@ -1,0 +1,35 @@
+import Control.Monad
+import Data.Foldable
+import System.FilePath
+import System.Directory
+import System.Process
+import System.Exit
+
+main :: IO ()
+main = do
+  let bootstrap_root = "./bootstrap-root"
+      happy_boot = bootstrap_root </> "happy"
+  callProcess "cabal"
+    [ "install", "happy",
+      "-f", "-bootstrap",
+      "--installdir=" ++ bootstrap_root ]
+  with_y_files "packages" $ \y_file -> do
+    putStrLn $ "Processing " ++ show y_file
+    callProcess happy_boot [y_file]
+    removeFile y_file
+  removePathForcibly bootstrap_root
+
+with_y_files :: FilePath -> (FilePath -> IO ()) -> IO ()
+with_y_files path cont = do
+  is_dir <- doesDirectoryExist path
+  if is_dir then do
+    entries <- listDirectory path
+    for_ entries $ \entry ->
+      with_y_files (path </> entry) cont
+  else do
+    is_file <- doesFileExist path
+    if is_file then do
+      let ext = takeExtension path
+      when (ext == ".y" || ext == ".ly") (cont path)
+    else do
+      die ("Neither file nor dir: " ++ show path)
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,6 @@
 # Revision history for Happy
 
-## Unreleased
+## 2.0
 
 The main focus of this release was modularizing Happy.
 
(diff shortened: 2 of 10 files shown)


bootstrap.hs
```diff
@@ -0,0 +1,35 @@
...
+import System.Process
+import System.Exit
+
+main :: IO ()
+main = do
+  let bootstrap_root = "./bootstrap-root"
+      happy_boot = bootstrap_root </> "happy"
+  callProcess "cabal"
+    [ "install", "happy",
+      "-f", "-bootstrap",
```
Ericson2314, line 14:
```suggestion
      "-f", "+bootstrap",
```
Don't we want that?

int-index: That would require a pre-installed `happy`

Ericson2314: No we don't

int-index: With `-f -bootstrap` we don't, with `-f +bootstrap` we do. That's how this flag works, and I think you named it 😄

Ericson2314: Sorry! I wasn't loading your comments, I meant to reply to myself not rudely reply to you! :flushed: :grimacing:

Ericson2314: Yes, this is not the first time my own name for the flag has confused me :sweat_smile:

andreasabel: I think this also confuses me.  
I would think that `-f +bootstrap` would do something so that _this_ instance of `happy` can build _itself_, whereas a `-f -bootstrap` would require _another_ (already installed) instance of `happy`.
Or in other words, _flag on_ does something special, _flag off_ does the vanilla thing.

andreasabel:
@int-index wrote:
> The new release process is:

These instructions shouldn't just reside in a closed PR, they need to go into `DEVELOPER.md` (or maybe `CONTRIBUTING.rst`).

Also the `bootstrap.hs` file should get a preamble with a purpose statement: What it achieves, when and how to invoke it.

PR #257, opened 2023-10-01
Reduce the amount of CPP for attribute grammar support

Ericson2314:
We previously had a lot of `#ifdef HAPPY_BOOTSTRAP` to support mangling with and without attribute grammar support. Now that module is broken up so that we need less.

This hopefully makes the code easier to understand and maintain.

### DIFF
--- a/packages/frontend/happy-frontend.cabal
+++ b/packages/frontend/happy-frontend.cabal
@@ -38,7 +38,9 @@ tested-with:
         GHC == 7.0.4
 
 flag bootstrap
-  description: Optimize the implementation of happy using a pre-built happy
+  description:
+    Optimize the implementation of happy using a pre-built happy,
+    and add support for attribute grammars.
   manual: True
   default: True
 
@@ -63,7 +65,7 @@ library
   other-modules:
         Happy.Frontend.ParseMonad
         Happy.Frontend.ParseMonad.Class
-        Happy.Frontend.AttrGrammar
+        Happy.Frontend.Mangler.Monad
         Happy.Frontend.Parser
         Happy.Frontend.Lexer
         Happy.Frontend.ParamRules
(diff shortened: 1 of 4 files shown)


packages/frontend/happy-frontend.cabal
```diff
@@ -38,7 +38,9 @@ tested-with:
         GHC == 7.0.4
 
 flag bootstrap
-  description: Optimize the implementation of happy using a pre-built happy
+  description:
+    Optimize the implementation of happy using a pre-built happy,
+    and add support for attribute grammars.
```
Ericson2314, line 43: Should mention the other behavioral thing this flag does.

packages/frontend/happy-frontend.cabal
```diff
@@ -63,7 +65,7 @@ library
   other-modules:
         Happy.Frontend.ParseMonad
         Happy.Frontend.ParseMonad.Class
-        Happy.Frontend.AttrGrammar
```
Ericson2314, line 66: This was actually already conditionally used, and should thus go below.

packages/frontend/happy-frontend.cabal
andreasabel: I would have the flag the other way round, so it would then be "Build a version of happy for boot-strapping.  The generated binary will not have support for attribute grammars."
See https://github.com/haskell/happy/pull/256#discussion_r1342398966.

int-index: It's going to be confusing either way because `bootstrap` can be interpreted either as "bootstrapped" or "for bootstrapping", which are opposite in meaning.

Any ideas for a less ambiguous name?

Ericson2314: Well that is a preexisting issue; I am happy to change it but seems like ti should be done separately. Does all the new things from this PR look good?

int-index (approved):

PR #267, opened 2024-01-24
Compile testsuite with -threaded (#264)

### LINKED ISSUE #264: Can't figure out how to run tests in parallel
The test.hs script calls make with `-j` set to the number of capabilities of the script process. I can't seem to figure out how to increase that number; at the least it does not pick up the number of cpu cores on my machine automatically.
It's a bit annoying, but not too bad: I can just change the script to use `-j7` (I have 6 cores). Nevertheless, the script seems a bit misleading.

sgraf812:
Fixes #264.


andreasabel (approved):
Thanks!

PR #279, opened 2024-06-17
Update CI configuration

int-index:
1. Dropped old GHC versions due to this error from `haskell-ci`:

    ```
    No haskell-ci.sh, skipping bash regeneration
    *ERROR* 'tested-with:' specifically refers to unknown 'GHC' versions: 7.0.4, 7.4.2, 7.6.3, 7.8.4, 7.10.3
    Known GHC versions: 8.0.1, 8.0.2, 8.2.1, 8.2.2, 8.4.1, 8.4.2, 8.4.3, 8.4.4, 8.6.1, 8.6.2, 8.6.3, 8.6.4, 8.6.5, 8.8.1, 8.8.2, 8.8.3, 8.8.4, 8.10.1, 8.10.2, 8.10.3, 8.10.4, 8.10.5, 8.10.6, 8.10.7, 9.0.1, 9.0.2, 9.2.1, 9.2.2, 9.2.3, 9.2.4, 9.2.5, 9.2.6, 9.2.7, 9.2.8, 9.4.1, 9.4.2, 9.4.3, 9.4.4, 9.4.5, 9.4.6, 9.4.7, 9.4.8, 9.6.1, 9.6.2, 9.6.3, 9.6.4, 9.6.5, 9.8.1, 9.8.2, 9.10.1
    ```

2. Run `haskell-ci regenerate`

The patch in `.github/haskell-ci.patch` doesn't apply cleanly, will look into it later.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.17.20231219
+# version: 0.19.20240608
 #
-# REGENDATA ("0.17.20231219",["github","cabal.project","--config","cabal.haskell-ci","--project"])
+# REGENDATA ("0.19.20240608",["github","cabal.project","--config","cabal.haskell-ci","--project"])
 #
 name: Haskell-CI
 on:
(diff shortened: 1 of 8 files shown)


sgraf812:
It appears we are affected by https://github.com/haskell-CI/haskell-ci/issues/738, so it would be good to update the generated workflows with `haskell-ci >= 0.19`.

phadej:
@sgraf812 I don't see how haskell-ci#738 is relevant. The failing runs are one month old and failing in `cabal test` step.

```
make: ../dist/build/happy/happy: No such file or directory
make: *** [Makefile:78: shift01.g.hs] Error 127
```

The `haskell-ci#738` would fail to run at all.

sgraf812:
I wasn't clear; the failures are over at https://github.com/haskell/happy/pull/281. We need to land this PR in order to fix the situation.

int-index:
I'm not working on this at the moment, so if someone wants to take over the branch, be my guest.

sgraf812:
Wow, this `[missing-bounds-important]` error is *absolutely* annoying. First, it doesn't tell you the line number or just the *file* of the violating dependency. And then it appears to even trigger for the **testsuite**. What on earth?!

It's a bit unfortunate we have to drop support for GHCs older than 8.2, but I don't know enough about the OS support to judge whether this is avoidable. (Perhaps hvr's PPA phased out with Ubuntu 22.04 and it's toomuch work to support GHC 7.0, say, in `ghcup`.)

sgraf812:
I think I'll just go ahead and merge this.

Perhaps I'll try to get rid of the complicated `.patch` setup afterwards.

PR #283, opened 2024-06-28
Happy catchtok

Kariiem:
This is a wip for introducing `catchTok` and all the neccessary generated code in the template files.


packages/backend-lalr/data/HappyTemplate.hs
```diff
@@ -194,35 +221,147 @@ happyDropStk n  (x `HappyStk` xs) = happyDropStk MINUS(n,(1#::Happy_Int)) xs
...
+--     When there is no `errorTok` action, call the error handler
+--     (e.g., `happyError`) with the resumption continuation `happyResume`.
+--  2. Error resumption mode: If the error handler wants to resume parsing in
+--     order to report multiple parse errors, it will call the resumption
+--     continuation (of result type `P (Maybe a)`).
+--     In the absence of the %resumptive declaration, this resumption will
+--     always (do a bit of work, and) `return Nothing`.
+--     In the presence of the %resumptive declaration, the grammar author
+--     can use the special `catch` terminal to declare where parsing should
+--     resume after an error.
```
sgraf812, line 244: This comment by me is out of date. Note that we no longer have a `%resumptive` pragma; we only have `%error { abort } { report }`, and `report` gets passed the resumption (of type `P a`).

The default behavior without `%error` (calling `happyError`) or with the unary variant of `%error` is to call a user-supplied function wrapped in `happyReport` that ignores the resumption. The reporting function will likely panic and abort the parse (at least `happyFail` does this). See also the code generation for `callReportError`.

It's all a bit confusing due to multiple layers of indirection; perhaps we can simplify.

It is probably best to describe the expected behavior of `%error { abort } { report }` in conjunction with the `catch` token in the user's guide first. Building on that documentation, we can describe how resumption is implemented here.

sgraf812:
Superseded by #318.

PR #284, opened 2024-07-15
Simplify CI workflow to standard `haskell-ci` workflow

sgraf812:
The simpler, the better.


sgraf812:
I'm not completely sure what changed, but now the vanilla CI workflow generated by `haskell-ci` just works.

It seems that our previous logic in `tests/Makefile` to set `HAPPY` was simply too smart for its own good.

Happy for @andreasabel to take a look. I verified that we indeed test the correct `happy` version which adding a dummy Makefile target `path.run` which outputs

```
--> Printing happy path...
which happy
/__w/happy/happy/dist-newstyle/build/x86_64-linux/ghc-8.6.5/happy-2.0/x/happy/build/happy/happy
```

andreasabel (approved):

sgraf812:
Thanks. I think your approach was necessary back when we still had the `+bootstrap` flag, but that was removed in #277.

PR #289, opened 2024-09-13
Use -XNoStrictData in generated parser (#273)

### LINKED ISSUE #273: Happy-generated code fails with `Internal Happy error` with `-XStrictData`
I am new to Happy and want to use it for an (old school) introductory course on compiler engineering. Because I repetitively got the very uninformative `Internal Happy error` error, I tested my whole setup with a minimal example. Surprisingly, it still fails with `Internal Happy error`.

Here is my setup:

```
module Main where

import Token
import Parser

main :: IO ()
main = do
  result <- parse [ VAR ]  
  print result -- should print 0
  return ()
```

```
module Token where

data Token = VAR deriving (Show, Eq)
```

```
{
module Parser where 

import System.IO
import Token
}

%name parse
%tokentype { Token }
%error { parseError }
%monad { IO } { >>= } { return }

%token
  VAR     { VAR }

%%

program : VAR { 0 }

{
parseError :: [Token] -> IO a
parseError _ = fail "something went wrong" -- I never see this error message ...
}
```

After diving a bit into the code I saw that `parse` invokes `happyParse start_state = happyNewToken start_state notHappyAtAll notHappyAtAll` with `start_state = 0#`. It seems like that somewhere deep inside the machinery the second `notHappyAtAll` is actually evaluated by Haskell, since it is accessed as a `HappyStk` here:

```
happyFail explist 0# tk old_st _ stk@(x `HappyStk` _) =
     let i = (case Happy_GHC_Exts.unsafeCoerce# x of { (Happy_GHC_Exts.I# (i)) -> i }) in
--      trace "failing" $ 
        happyError_ explist i tk
```

As far as I understand, this means that Happy fails with an empty stack - probably right before even reading the first token. 

Here is some information about my environment:
- Late 2023 Mac with the `M3` apple silicon
- `ghc --version` prints `9.4.8`
- `cabal --version` prints `3.10.2.1`
- `happy --version` prints `1.20.1.1`

My `build.cabal`:
```
cabal-version:      3.0
build-type:         Simple

...

executable HopefullyHappySoon:
  hs-source-dirs:     src
  main-is:            Main.hs
  other-modules:      Token,
                      Parser
  build-depends:      base ^>= 4.17.2.1, 
                      array >= 0.5.4.0
  build-tool-depends: happy:happy >= 1.20.1.1 -- Just to be sure ...
  default-language:   GHC2021
```

I appreciate any kind of help in this situation. This issue is also related to #241.

sgraf812:
Fixes #273. CC @andreasabel

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,10 @@
 
 The main focus of this release was modularizing Happy.
 
+* Generated parsers now activate the language extension `-XNoStrictData` without
+  which every use of a happy parser would lead to an immediate crash (#273).
+  This causes us to drop support for GHC < 8.0.
+
 * Break up into libraries:
 
   - `happy-backend-glr`
--- a/happy.cabal
+++ b/happy.cabal
@@ -140,7 +140,7 @@ executable happy
   hs-source-dirs: src
   main-is: Main.lhs
 
-  build-depends: base < 5,
+  build-depends: base >= 4.9 && < 5,
                  array,
                  containers >= 0.4.2,
                  mtl >= 2.2.1,
@@ -164,5 +164,5 @@ test-suite tests
   build-tools: happy
   ghc-options: -threaded
 
-  build-depends: base < 5, process < 1.7
+  build-depends: base >= 4.9 && < 5, process < 1.7
   default-language: Haskell98
--- a/packages/backend-lalr/src/Happy/Backend/LALR.hs
+++ b/packages/backend-lalr/src/Happy/Backend/LALR.hs
@@ -35,7 +35,7 @@ importsToInject debug = concat ["\n", import_array, import_bits, import_glaexts,
                            "import Control.Monad (ap)\n"
 
 langExtsToInject :: [String]
-langExtsToInject = ["MagicHash", "BangPatterns", "TypeSynonymInstances", "FlexibleInstances"]
+langExtsToInject = ["MagicHash", "BangPatterns", "TypeSynonymInstances", "FlexibleInstances", "NoStrictData"]
 
 defines :: Bool -> Bool -> String
 defines debug coerce = unlines [ "#define " ++ d ++ " 1" | d <- vars_to_define ]


ChangeLog.md
```diff
@@ -4,6 +4,9 @@
 
 The main focus of this release was modularizing Happy.
 
+* Generated parsers now activate the language extension `-XNoStrictData` without
```
phadej, line 7: `-X(No)StrictData` is available since GHC-8.0, so this change means that `happy` won't anymore generate code which works with older GHCs.

As far as I'm concerned, that is fine, but it's probably a good idea to tell explicitly in the changelog.

phadej: ... you don't test `master` with older than GHC-8.0, though it might be still buildable with older GHCs? If it weren't, than using new happy with old GHCs will become very unlikely.

sgraf812: The trouble is that it has become quite hard to test 7.10 and below in CI. `ghcup` doesn't provide it and I'm not a huge fan of manually setting it up. Besides it is not unlikely that it won't work due to some glibc incompatibility.

Besides, GHC 8.0 is quite ancient at this point and I would be happy to drop support for 7.10 and below. The cabal file says as much already, but perhaps we should announce it in the release notes as well.

int-index (approved):
I've had to insert `{-# LANGUAGE NoStrictData #-}` manually on multiple occasions, so I support fixing this in the code generator.

ChangeLog.md
phadej: > The cabal file says as much already,

**It doesn't**

It only says that you test with GHC-8.0 and up, not that you don't support oldere GHCs. E.g. you have

```
  build-depends: base < 5,
                 array,
                 ...
```

so technically cabal would try compile even with GHC-7.0 etc.

And it actually fails, if you test it:

```
Building library for happy-backend-glr-2.0..
ghc: unrecognised flag: -Wincomplete-uni-patterns
```

phadej: @int-index Correct the metadata. Write lower-bounds. `base >=4.9`. Not a hard thing to do. It's not a waste of time.

int-index: Dependency bounds for the generator itself and dependency bounds for the generated code aren't the same thing. `base >=4.9` is correct as long as `happy` itself builds with `base >=4.9`.

I agree that we should bump this bound because, as you point out, `happy-backend-glr` doesn't build with older GHCs, however it's a separate problem unrelated to this PR.

phadej: As I said, if `happy` itself is not buildable with older GHCs then it's make sense it isn't able to generate code for older GHCs.

It could be so that `happy` is buildable only with newer GHCs (say even GHC-9+ only) but supports code generation for older GHCs. (It would be trickier to test but not impossible. E.g. `cabal` is buildable with only newer GHCs, but can work with older than  that).

It would be somewhat confusing if `happy` would be buildable with old GHCs, but not be able to generate code for those old GHCs. Given that `cabal-install` builds `happy` itself (doesn't reuse some global installation), those users would find themselves if complicated situations.

So, IMHO, it makes everything somewhat simpler if `happy` itself and generated code has the same support GHC windows. *"If you can build a `happy` version with GHC-x.y, that `happy` will generate code which GHC-x.y will handle"*.

In particular, if/when GHC introduces a change, such that `happy` itself is still buildable, but code is not, I'd argue that `happy` shouldn't be buildable.

**There is no way to declare in `happy.cabal` dependency metadata for the generated code**, so at large it's simpler if `happy`'s own build metadata is (re)used also as the generated code dependency metadata. (Similarly as we are currently imprecise in multi-staged or cross-compilation scenarios, using the same bounds for all stages and host&target).

sgraf812:
Alright, I added another commit setting a lower bound on `base >= 4.9`, simply because we discussed the issue here. I will merge soon.

PR #291, opened 2024-09-13
Generate two spaces instead of tabs (#24)

### LINKED ISSUE #24: Stop using tabs in generated parsers
The title pretty much speaks for itself—using tabs for layout is usually not the greatest idea, so Happy shouldn't be producing code that does so. The simplest way to do this is probably to replace `\t` with eight literal spaces everywhere it appears in the code generator source. The clean way is probably to add a function `tabIn = ("        "++)`

sgraf812:
Fixes #24. Do we want to?


int-index (approved):
Let's do this. I see no reason at all to emit those tabs.

sgraf812:
Alright, I'll just merge before someone starts a fight around tabs vs. spaces.

Issue #288, opened 2024-09-13
Should we make the `happy-*` packages public library components of `happy`?


sgraf812:
Cabal 3.0 introduced the [`library:visibility` field](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-library-visibility).
Should we use it to define the single-library-component `happy-*` packages as "sublibraries"? Depending on such a library is exemplified here: https://github.com/haskell/cabal/issues/9480#issuecomment-1826883582. For us it would lead to names such as `happy:tabular`.
Defining such public sublibraries is exemplified here: https://github.com/BlockScope/plugins-for-blobs/blob/develop/plugins-for-blobs.cabal. Note the `visibility: public` fields.

Pros:

* No separate .cabal file, change log, package description, supported versions, etc. for each package to maintain. I think this is huge, because maintainer time is scarce.

Cons:
 
* Needs Cabal 3.0 (released in 2019). 

Given that

1. we already decided to lock into the GHC ecosystem 
2. `cabal-install` is easily buildable with older GHCs
3. older `happy` versions will continue to build with ancient GHCs and accept the same .y-file syntax,

I don't think requiring Cabal 3.0 is much of a drawback. On the other hand I *do* see the real costs associated with release management if we stick to one executable package and 5 library packages, all of which need to keep their versions in sync etc.

**This decision has potential to lock us in for quite a long time, as `happy-tabular` is a library name (technically a package name which implicitly denotes the default library) than `happy:tabular`. I would like to have the Go from @int-index, @Ericson2314 and @andreasabel on this.**

int-index:
How well do `stack` and `nix` support `visibility: public`?

Since you point out the difference between `happy-tabular` and `happy:tabular`, how would `nixpkgs` deal with this? What do we get instead of `haskellPackages.happy-tabular`?

int-index:
Here are the only packages I could find on Hackage that use `visibility: public` in their `.cabal` files:

* https://hackage.haskell.org/package/dani-servant-lucid2
* https://hackage.haskell.org/package/pact-time
* https://hackage.haskell.org/package/red-black-record
* https://hackage.haskell.org/package/saturn
* https://hackage.haskell.org/package/spotify
* https://hackage.haskell.org/package/swarm
* https://hackage.haskell.org/package/xnobar

Out of these packages, I found only `saturn` on Stackage:
* https://www.stackage.org/nightly-2024-09-13/package/saturn-1.0.0.5

[...]

PR #293, opened 2024-09-13
Merge `happy-*` library packages into a single `happy` library component (#292)

### LINKED ISSUE #292: Merge `happy-*` library packages into a single `happy` library component
Triggered by the discussion in https://github.com/haskell/happy/issues/288, we found that it would be preferable to merge all `happy-*` library packages into the library component associated with the `happy` executable. The reasons are:

* Less maintenance overhead compared to the status quo (see #288).
* `happy` does not have many dependencies, so there is almost no added value of splitting it into several libraries. For example, `happy-backend-lalr` does not take much longer to compile than `happy-grammar`. 
* Multiple public library components as in Cabal 3.0 allow to retain the split, ~~but none of the compilation time benefits apply because [all components are built anyway](https://github.com/haskell/happy/issues/288#issuecomment-2348942466)~~ Edit: wrong according to @phadej; see below. However, multiple public components might also be less well supported by `stack` and `nix`.

I will prepare a PR.

sgraf812:
See #292. Fixes #292.

I'm delighted that git is able to map the renamings, which should make this PR easy to review.


Ericson2314 (changes requested):
I really do not like this. If the modular boundaries are not enforced, everything will bleed back together. Humans cannot be trusted to manually enforce boundaries during code review, the computer must help.

Does multi-repl not solve all the annoyances of having multiple libraries?

int-index:
This is a good change because it removes lots of administrative overhead. @Ericson2314's concern is valid, but given the lack of resources to maintain `happy` I find it justified to prioritize ease of maintenance, release management, etc, over enforcing modularity boundaries.

@Bodigrim's suggestion in https://github.com/haskell/happy/issues/288#issuecomment-2349880724 could be a viable alternative, too. Either way, we need to get rid of those separate `.cabal` files.

sgraf812:
Closed in favor of #297.

PR #297, opened 2024-09-14
New package `happy-lib`; make previous `happy-*` packages sublibraries (#288)

### LINKED ISSUE #288: Should we make the `happy-*` packages public library components of `happy`?

sgraf812:
This is one design design discussed in #288.

Fixes #288.


lib/happy-lib.cabal
```diff
@@ -0,0 +1,168 @@
...
+        Happy.Frontend.ParseMonad.Class
+        Happy.Frontend.Mangler.Monad
+        Happy.Frontend.Parser
+        Happy.Frontend.Lexer
+        Happy.Frontend.ParamRules
+        Happy.Frontend.AttrGrammar
+        Happy.Frontend.AttrGrammar.Parser
+        Happy.Frontend.AttrGrammar.Mangler
+
+library tabular
```
int-index, line 102: What's the relation between this `library tabular` subsection and `lib/tabular/happy-tabular.cabal`? I didn't expect to see the latter.

Ericson2314: I think we can just remove the others?

Ericson2314 (commented):
Same question as @int-index about old cabal files, but very big +1 on the design!

lib/happy-lib.cabal
sgraf812: Good point, I forgot to remove the .cabal files. The main thing about this approach :)

Should be done now.

int-index (approved):

sgraf812:
Alright then, let's do it.

PR #298, opened 2024-09-16
Attribute grammar cleanup

Ericson2314:
I want to better separate the attribute grammar stuff

- Much code doesn't care at all (rightly or wrongly), so we can use the types to track which does does.

- Only one backend supports it, it should be ill-typed to try to use it with the other backend. (This is not yet achieved, but step in that directly.

- Generally trying to slim down `Grammar` to just the essentials, the grammar itself.

### DIFF
--- a/app/Main.lhs
+++ b/app/Main.lhs
@@ -82,7 +82,7 @@ Parse, using bootstrapping parser.
 
 Mangle the syntax into something useful.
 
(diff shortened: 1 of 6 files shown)


Ericson2314:
Anyone want to review?

sgraf812:
LGTM!

PR #300, opened 2024-09-16
Grammar of

Ericson2314:
Instead of assuming we always have raw Haskell code (a `String`) which
eliminates each production rule, use a type variable.

As one would hope, much code only cares about the grammar itself, and
works with that type abstract. The new types thus recover free theorems,
yay! Only a little bit of the frontend and the backends actually wants
it to be strings.

These type parameter can be leveraged in the future. Some possibilities
are:

- a TH-based frontend to Happy

- interpreting the grammar where we don't have custom elimination rules
  at all, but just create an "untyped" parse tree.

Depends on #298


lib/frontend/src/Happy/Frontend/Mangler.lhs
```diff
@@ -27,32 +29,35 @@ Mangler converts AbsSyn to Grammar
...
+>   .  e
+>   -- ^ Empty elimination rule, used for starting productions. Will never be run.
+>   -> CodeChecker e
 >   -> FilePath
->   -> AbsSyn
->   -> M (Grammar, Pragmas)
-> manglerM checkCode file (AbsSyn dirs rules') =
+>   -> AbsSyn e
+>   -> M (Grammar e, Pragmas)
+> manglerM noCode checkCode file (AbsSyn dirs rules') =
```
Ericson2314, line 60: My favorite part! Nice preservation of `e` through the higher-order function

Ericson2314:
OK this is ready for review too!

sgraf812:
LGTM! It's a nice change, as you say.

PR #302, opened 2024-09-19
Better panic message, indicating that it's not the user's fault (#186)

### LINKED ISSUE #186: Use more descriptive error in case of runtime failure
`notHappyAtAll` doesn't provide any useful information, not even stack trace, so user is left wondering what happened in generated file.

sgraf812:
There were enough requests for a better panic message that we may as well deliver it.

Fixes #186.


int-index:
Good idea

Ericson2314 (approved):

Issue #271, opened 2024-01-26
`happyExpToks` table can be larger than `happyTable`


sgraf812:
For tests/issue93.y, the `happyExpToks` array is about 20% larger than `happyTable`. We should definitely try to compress it more, perhaps use an array of lists. Better: Error handling is not required to be fast; just reconstruct the expected tokens from the `happyTable`.

int-index:
>  Error handling is not required to be fast

Are you sure? GHC relies on the `error` token to implement the layout rule.

sgraf812:
Yes, in https://github.com/haskell/happy/pull/272 I produce the expected tokens lazily when the error handler (`reportError`, in the example) needs them. For the `error` token, this error handler will not even be called, unless there really is a syntax error.

PR #272, opened 2024-02-02
Draft: Introduce new `%error` handler mode and `catch` for resumable parsing

sgraf812:
Superseded by #318.

A drafty PoC serving as motivation for a GSoC proposal; I don't want to see this merged or reviewed for now. 

---

Consider this excerpt from an example (`errormonad-resume.y`):

```hs
%name parseStmts Stmts
%tokentype { Token }
%error { \_ -> abort } { reportError } -- the entire point of this test

%monad { ParseM } { (>>=) } { return }

%token
  '1' { TOne }
  '+' { TPlus }
  ';' { TSemi }

%%

Stmts : {- empty -}           { [] }
      | Stmt                  { [$1] }
      | Stmts ';' Stmt        { $1 ++ [$3] }
      | catch ';' Stmt %shift { [$3] } -- Could insert error AST token here in place of $1

Stmt : Exp { ExpStmt $1 }

Exp : '1'                { One }
    | Exp '+' Exp %shift { Plus $1 $3 }

{
recordParseError :: [String] -> ParseM ()
recordParseError expected = recordError [ParseError expected]

reportError :: [Token] -> [String] -> ([Token] -> ParseM a) -> ParseM a
reportError ts expected resume = do
  recordParseError expected
  resume ts
```

The point of this example is that *reporting* a parse error (i.e. adding a diagnostic) is independent from *aborting* a parse (i.e., a fatal error that can't produce a syntax tree). Hence the new `%error` form takes two code blocks: One for the abort handler, the other for the report handler.

Additionally, the report handler gets a `resume` continuation that it may call to resume parsing; otherwise it would simply have to abort (TODO: In the current encoding it should perhaps simply be `ParseM ()` and we call the resumption unconditionally).

Where does the parser resume parsing? That is mostly up to the user to specify, through use of the special `catch` terminal. In the example above, `catch` occurs before a `;` is shifted, so that upon an error during parsing a `Stmt`, the parser will "unwind" the stack until it finds a situation in which `catch` can be shifted. After having done that, it will discard input until it finds the next `;` so as to resume parsing.

The result is that for inputs such as `1++1;1;+`, two errors (and a partial syntax tree) can be reported: One at the second `+` and the other at the third.

I've already tried to apply this patch to GHC; it seems to work: See https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11990 for a worked example.

PR #318, opened 2024-10-07
Introduce `catch` and new `%error` handler mode for resumable parsing 

sgraf812:
This is the "production-ready" version of #272 and #283, superseding both.

---

This patch enables users of happy to emit multiple parse errors in the same file and produce a partial syntax tree.

* The main feature is described in the user's guide entry. 
* I also added a user's guide entry for the hitherto undocumented "expected token list" feature (which I broke to vastly improve it, but it was not documented anywhere). This will fix https://github.com/haskell/happy/issues/271 because we no longer need an uncompressable table.
* The most important place documenting the implementation and the backwards compatibility dance with the existing `error` token is described in `Note [Error recovery]` and its children.

This patch unblocks our GSoC student @Kariiem's work over at https://gitlab.haskell.org/ghc/ghc/-/merge_requests/13145.


Ericson2314:
This is very cool! :)

lib/tabular/src/Happy/Tabular/First.lhs
```diff
@@ -38,23 +37,25 @@ This will never terminate.
 >                  , lookupProdsOfName = prodsOfName
 >                  , non_terminals = nts
 >                  })
->       = joinSymSets (\ h -> maybe (Set.singleton h) id (lookup h env) )
+>       = joinSymSets (\h ->fromMaybe (Set.singleton h) (lookup h env))
```
Ericson2314, line 40:
```suggestion
>       = joinSymSets (\h -> fromMaybe (Set.singleton h) (lookup h env))
```

sgraf812:
Will merge soon. It would be good if anyone could give the user's guide entry a read and tell me if it made any sense.

Ericson2314:
The users guide made sense, that was the only part I read.

PR #321, opened 2024-10-19
Qualify every use of Prelude functions (#131)

### LINKED ISSUE #131: qualify Prelude functions in templates?

sgraf812:
This bit end users in visq/language-c#99.

Properly fixes #131 and adds a regression test.


Bodigrim:
This is a great improvement, but it might be nice to document somewhere that `happy` relies on `Prelude` being `base:Prelude`, because in general it can be overridden in Cabal file with something like `mixins: base hiding (Prelude), relude (Relude as Prelude)`.

(I'm not sure whether it's possible or practical to resolve the general issue)

sgraf812:
I had hoped to import Prelude qualified as `Happy_Pre` and then use a package import, but that means that the implicit `import Prelude` is no longer inserted (Section 5.6.1 of the Haskell report). 

So happy cannot by itself add a qualified import of Prelude without surprising effect on user programs.
It appears that users who `import Prelude (...)` or redefine `Prelude` must also `import qualified "base" Prelude`. It's a bit ugly, but not too complicated to detect and adhere to.

I documented this in the user's guide.

andreasabel:
This PR broke Agda: 
- https://github.com/agda/agda/issues/7585

Because our parser has `import Prelude hiding (null)` another `import qualified Prelude` has to be omitted by the generator to have `Prelude.null` in scope.

I think 2.1.1 should be deprecated as it pretends to be a minor bump but breaks user's code.

I'll open a new issue for all that, no need to reply here.

PR #322, opened 2024-10-19
Resurrect the `%errorhandlertype` directive for back-compat (#320)

### LINKED ISSUE #320: version 2.1: unrecognised directive: %errorhandlertype
It appears that as of version 2.1 `happy` reports:
```
unrecognised directive: %errorhandlertype
```
This directive is still in the docs, so if it is removed indeed, it'd be nice to update the docs to say what changed, and what to do instead.

sgraf812:
Fixes #320.


Bodigrim:
FWIW `happy-2.1` was a major bump already, which is [enough for a breaking change](https://pvp.haskell.org/#version-numbers) (even if not reflected in the changelog).

sgraf812:
Thanks. Indeed I was wondering whether `happy`'s versioning scheme is subject to the PVP (it appears that every package is). I bumped to 2.1.1 instead.

sgraf812:
@yav I verified that the package candidate 2.1.1 makes `language-rust` buildable again. I'll merge soon.

PR #326, opened 2024-10-26
Do not use Prelude; concoct our own Happy_Prelude alias (#325)

### LINKED ISSUE #325: The qualified use of `Prelude` introduced in 2.1.1 breaks Agda
Happy 2.1.1 broke Agda: 
- https://github.com/agda/agda/issues/7585

Because our parser has `import Prelude hiding (null)` another `import qualified Prelude` has to be omitted by the generator to have `Prelude.null` in scope.

I think 2.1.1 should be deprecated as it pretends to be a minor bump but breaks user's code.

A new happy should be released including the fix: Emit `import qualified Prelude` in the generated code.

sgraf812:
This is in order to undo a breaking change caused by users writing `import Prelude hiding (null)`.

Fixes #325.

### DIFF
--- a/lib/backend-lalr/src/Happy/Backend/LALR.hs
+++ b/lib/backend-lalr/src/Happy/Backend/LALR.hs
@@ -19,7 +19,7 @@ magicFilter magicName = case magicName of
       in filter_output
 
 importsToInject :: Bool -> String
-importsToInject debug = concat ["\n", import_array, import_list, import_bits, import_glaexts, debug_imports, applicative_imports]
+importsToInject debug = concat ["\n", import_prelude, import_array, import_bits, import_glaexts, debug_imports, applicative_imports]
     where
       debug_imports  | debug     = import_debug
                      | otherwise = ""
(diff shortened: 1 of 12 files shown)


sgraf812:
cabal.project to test the candidate:

```
packages:
  . 
  http://hackage.haskell.org/package/happy-lib-2.1.2/candidate/happy-lib-2.1.2.tar.gz
  http://hackage.haskell.org/package/happy-2.1.2/candidate/happy-2.1.2.tar.gz
```

lib/backend-lalr/src/Happy/Backend/LALR.hs
```diff
@@ -19,16 +19,16 @@ magicFilter magicName = case magicName of
...
-importsToInject debug = concat ["\n", import_array, import_list, import_bits, import_glaexts, debug_imports, applicative_imports]
+importsToInject debug = concat ["\n", import_prelude, import_array, import_bits, import_glaexts, debug_imports, applicative_imports]
     where
       debug_imports  | debug     = import_debug
                      | otherwise = ""
       applicative_imports        = import_applicative
 
       import_glaexts     = "import qualified GHC.Exts as Happy_GHC_Exts\n"
       _import_ghcstack   = "import qualified GHC.Stack as Happy_GHC_Stack\n"
+      import_prelude     = "import qualified Data.Bool as Happy_Prelude\nimport qualified Data.Function as Happy_Prelude\nimport qualified Data.Maybe as Happy_Prelude\nimport qualified Data.Int as Happy_Prelude\nimport qualified Data.String as Happy_Prelude\nimport qualified Data.List as Happy_Prelude\nimport qualified Control.Monad as Happy_Prelude\nimport qualified Text.Show as Happy_Prelude\nimport qualified GHC.Num as Happy_Prelude\nimport qualified GHC.Err as Happy_Prelude\n"
```
andreasabel, line 30: Wouldn't this be more readable as ?
```haskell
    import_prelude = unlines $ map (\ x -> unwords ["import qualified", x, "as Happy_Prelude"]) $ 
      [ "Data.Bool"
      , "Data.Function"
      , ... 
      ]
```

andreasabel (commented):
Would `HappyPrelude` (CamlCase) be more in style of this codebase?

andreasabel:
> cabal.project to test the candidate:
> 
> ```
> packages:
>   . 
>   http://hackage.haskell.org/package/happy-lib-2.1.2/candidate/happy-lib-2.1.2.tar.gz
>   http://hackage.haskell.org/package/happy-2.1.2/candidate/happy-2.1.2.tar.gz
> ```

I tested this with building Agda-2.6.4.3, and it worked fine!

sgraf812:
> Would HappyPrelude (CamlCase) be more in style of this codebase?

Perhaps, yes. On the other hand, all the other qualified imports are of the form `Happy_Data_Array` (for example). Furthermore, there is an alias `type Happy_Int = Int#`, so it's not like there is a fixed convention to follow.

I think I would prefer `HappyPrelude`, but if we switch to that case convention, we should do it consistently throughout the generated code. Not in this patch, though, and it is not an itch that I feel much of an urge to scratch.

PR #327, opened 2024-11-06
Auto-resolve shift/reduce conflicts involving the catch token

sgraf812:
When the current token is the catch token, we never reduce; we only ever shift (see the [user's guide](https://haskell-happy.readthedocs.io/en/latest/using.html#resumptive-parsing-with-catch) for how `catch` works). More precisely, the mechanism considers multiple situations in which the `catch` token can be shifted in parallel, including situations that are reachable via reductions. Hence it does not make sense to report shift/reduce conflicts involving the catch token: Error resumption mode will only ever try to shift it.

The solution implemented in this patch is not to generate conflicting `LR'Reduce` actions in the first place by deleting the `catch` token from the lookahead sets of LR1 items.

### DIFF
--- a/lib/grammar/src/Happy/Grammar.lhs
+++ b/lib/grammar/src/Happy/Grammar.lhs
@@ -175,6 +175,10 @@ For array-based parsers, see the note in Tabular/LALR.lhs.
 > catchName = "catch"
 > dummyName = "%dummy"  -- shouldn't occur in the grammar anywhere
 
+TODO: Should rename firstStartTok to firstStartName!
+It denotes the *Name* of the first start non-terminal and semantically has
+nothing to do with Tokens at all.
+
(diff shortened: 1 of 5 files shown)


lib/grammar/src/Happy/Grammar.lhs
```diff
@@ -175,6 +175,10 @@ For array-based parsers, see the note in Tabular/LALR.lhs.
 > catchName = "catch"
 > dummyName = "%dummy"  -- shouldn't occur in the grammar anywhere
 
+TODO: Should rename firstStartTok to firstStartName!
+It denotes the *Name* of the first start non-terminal and semantically has
+nothing to do with Tokens at all.
```
Ericson2314, line 180: Glad the newtypes are helping :)

PR #342, opened 2025-08-05
Add OPTIONS_HAPPY pragma for passing flags to happy

TOTBWF:
It is surprisingly difficult to get haskell build tools to pass options to `happy`: `cabal` only lets you do it via `cabal.project` files or via `--happy-options`, and the `--happy-option` flag in `stack` is just flat-out broken. This PR tries to fix this by adding a `OPTIONS_HAPPY` pragma to the `happy` grammar, which lets you specify options to pass to happy directly within the `.y` file.

### DIFF
--- a/happy.cabal
+++ b/happy.cabal
@@ -114,6 +114,7 @@ extra-source-files:
         tests/monad001.y
         tests/monad002.ly
         tests/monaderror.y
+        tests/Pragma.y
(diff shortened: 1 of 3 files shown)


sgraf812 (commented):
Thanks for doing this! I agree it would be a nice feature to have. But I have a few smaller suggestions.

PR #353, opened 2026-03-17
ISSUE 352: Fix extra initial LR(0) seed in genLR0items

marco-comini:
Fixes #352.

This patch fixes an off-by-one bug in `genLR0items`.

Previously, the initial LR(0) kernels were seeded with:

```haskell
[0 .. n_starts]
```

where `n_starts = length (starts g)`.

This caused one extra initial seeded component to be generated. Consequently, Happy could introduce many additional states coming from the first ordinary grammar production, even when that production was unrelated to the declared entry point.

This patch changes the seed range so that only the intended start rules are used.

I also added a regression test (`tests/issue352.y`).


sgraf812:
@andreasabel and @marco-comini if you have the time, give https://hackage.haskell.org/package/happy-lib-2.2.1/candidate and https://hackage.haskell.org/package/happy-2.2.1/candidate a short glance to see if everything looks correct. Will release tomorrow

